PR #2, opened 2017-12-07
Fix typo

olimart:


mikker:
Look at me trying to be fancy with french without knowing french 🙄

Thanks!

olimart:
Don't worry I also had to correct DHH once 😃

PR #3, opened 2017-12-08
before_filter is deprecated

koppen:
As of Rails 5.1(?) using `before_filter` will result in

> undefined method `before_filter' for ... Did you mean? before_action`

and a deprecation warning in earlier versions of Rails.


mikker:
How could I forget?! Thanks Jakob.
❤️🧡💛💚💙💜🖤

PR #5, opened 2017-12-14
Api docs (and rubocop)

mikker:


mftaff:
I have started working on adding documentation. I just want to make sure that the style is aligned with what you are looking for :-) 

Here is a link to the [api-docs] (https://github.com/mftaff/passwordless/commit/f24f6201bc269ba5a7110887f6e43dc9e7cb347d) branch on my fork. If/When  you have a minute to  look it over, your feedback would be amazing.
I want my contributions to be helpful to the project, so if you don't like any/part/all of it, please let me know, so I can learn better what the end goal is!

Again Thanks for letting me contribute!

mftaff:
What does this method return? I want to make sure I understand what `authenticatable_class` refers to.

    def authenticatable
      params.fetch(:authenticatable)
    end

Thanks!

mikker:
@mftaff It comes from this: https://github.com/mikker/passwordless/blob/master/lib/passwordless/router_helpers.rb#L8

It gets defined when the user does `passwordless_for :users` in her routes file. Eg. `:users` => `"user"`. I'm using `.fetch` to make sure that we raise an exception if it isn't found. `params[:authenticatable]` would just be `nil` and make trouble some place else 🙄 😄

mftaff:
@mikker A bit of a technical question... 
How do I push my changes to add them to this pull request? I have never done this before, and I couldn't find a clear answer on google 😛 
Do I push straight to this branch from my fork? Or do I create a PR from my fork? Or something else entirely?

mikker:
The easiest way is probably to create a pull request from your fork. It's based off of this branch, right? Then we can just merge that and delete this PR.

mikker:
Merged in https://github.com/mikker/passwordless/pull/7

PR #6, opened 2017-12-18
add YARD docs for all files in lib/ and app/

mftaff:


mikker:
You can check all those Hound warnings locally using `$ rubocop` in the project dir :)

mftaff:
Awesome, good to know 😄 
I will go fix 'em now.

lib/passwordless/url_safe_base_64_generator.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 module Passwordless
+  # Generates secure random numbers for sessions/cookies, etc.
   class UrlSafeBase64Generator
+    # Generates a url safe base64 secure random number. :-)
+    # @param _session [Object] some Object that contains this class.
```
mftaff: @mikker  I wasn't sure if I fully understood how to use this param. Maybe there is a better way to describe it?

lib/passwordless/router_helpers.rb
```diff
@@ -1,7 +1,14 @@
 # frozen_string_literal: true
 
 module Passwordless
+  # Some helpers for generating passwordless routes.
   module RouterHelpers
+    # Generates passwordless routes for any Model connected to Passwordless.
+    # @param resource [Symbol] the pluralized symbol of a Model (e.g - :users).
+    # @param at [string] Optional - provide custom path (i.e controller)
```
mftaff: @mikker I also wasn't sure if I understood what could be provided for `at` and `as`.
Can you review these lines as well?

lib/passwordless/url_safe_base_64_generator.rb
mikker: We pass the session to the generator so that if you wanted to use it in any way in your custom token generator you can. Say your user has a name you could make a custom generator that uses that for the tokens:

```ruby
class NameTokenGenerator
  def self.call(session)
    "#{session.authenticatable.name}-gets-a-token"
  end
end
```

Happens here:
https://github.com/mikker/passwordless/blob/master/app/models/passwordless/session.rb#L25

lib/passwordless/router_helpers.rb
mikker: Perhaps it's best shown with examples:

```
passwordless_for :users, at: 'session_stuff', as: :user_session_things
```

The passwordless engine get's mounted at `at`, eg your urls end up like:

```
/session_stuff/sign_in
```

Your url helpers get scoped under `as`, eg in a view:

```erb
<%= link_to 'Sign in', user_session_things.sign_in_path %>
```

mftaff:
@mikker Okay, I made those changes 😃 thanks for clarifying
How does it look now?

mikker:
Merged in https://github.com/mikker/passwordless/pull/7
Thank you for your help! 💙💚💛💜❤️

mftaff:
😃 If I can ever be of service again, please feel free to reach out.

mikker:
Sure, how about this: https://github.com/mikker/passwordless/issues/8 😃 
What do you think?

Issue #8, opened 2017-12-27
Make sessions (actually) expire


mikker:
Sessions don't expire right now. They get `expires_at` set to a year in the future but we don't really use that anywhere. It's meant to make the cookies expire at some point, forcing the user to sign in again. Probably a good idea.

(NB: Expiry is different than timeout. Timeout is _sign in using this token **before**_. Expiry is _sign user out automatically **after**_)

Now, we could do this by setting an expiry on the cookie itself. But I'm not sure. I think I'd like to check it programmatically, so we can alert the user of the reason they need to sign in again, e.g _Your session has expired, please sign in again_. I think the best place to do this for now is in the sessions controller.

mftaff:
Sure! I am keen to work on this.

mftaff:
Hear are my thoughts so far:

in sessions_controller.rb
```
...
def show
     ...
      session = find_session
      sign_in session.authenticatable
    ...
end
...
private
...
def find_session
    Session.valid.find_by!(
      authenticatable_type: authenticatable_classname,
      token: params[:token]
    )
end
...
```
In `find_session` you used `.valid` to scope to sessions that are not timed-out or expired. We probably want to remove that scope, (or at least remove the expires_at part of it.) in order that we can run a manual check for expires_at after finding a session? I think the solution could look something like this:

```
...
def show
     ...
      session = find_session
      sign_in session.authenticatable unless session_expired session
    ...
end
...
private
...
def find_session
    Session.valid.find_by!(
      authenticatable_type: authenticatable_classname,
      token: params[:token]
    )
end

def session_expired(session)
    if session.expires_at > Time.current
         false # i.e the session hasn't expired.
    else
        # add message saying session expired. using flash?
        true # i.e the session has expired.
    end
end
...
```

What do you think? Is this what you had in mind?

mftaff:
@mikker hey man, hope the new year is going well :-) I am curious if you had a chance to look this over? Do you want me to submit a PR?

mikker:
Yes, sorry, thanks! A PR is very welcome. Good thoughts so far. I have a few suggestions but the best way forward is maybe you getting as far as you can and then we work our way to the finish line together, bit by bit?

mikker:
Fixed in https://github.com/mikker/passwordless/pull/10

PR #10, opened 2018-01-14
Expire sessions

mftaff:
Relevant to issue #8

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -5,6 +5,9 @@ require 'bcrypt'
 module Passwordless
   # Controller for managing Passwordless sessions
   class SessionsController < ApplicationController
+    # Raise this exception when a session is expired.
+    class ExpiredSessionError < StandardError; end
+
     include ControllerHelpers
 
     helper_method :authenticatable_resource
@@ -31,6 +34,8 @@ module Passwordless
       render
     end
 
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+
     # get '/sign_in/:token'
     #   Looks up session record by provided token. Signs in user if a match
     #   is found. Redirects to either the user's original destination
@@ -42,6 +47,8 @@ module Passwordless
       BCrypt::Password.create(params[:token])
 
       session = find_session
+      raise ExpiredSessionError if session.expired?
+
       sign_in session.authenticatable
 
       redirect_enabled = Passwordless.redirect_back_after_sign_in
@@ -52,7 +59,11 @@ module Passwordless
       else
         redirect_to main_app.root_path
       end
+    rescue ExpiredSessionError
+      flash[:error] = 'Your session has expired, please sign in again.'
+      redirect_to main_app.root_path
     end
+    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
     # match '/sign_out', via: %i[get delete].
     #   Signs user out. Redirects to root_path
--- a/test/models/passwordless/session_test.rb
+++ b/test/models/passwordless/session_test.rb
@@ -16,7 +16,6 @@ module Passwordless
 
     test 'scope: valid' do
       valid = create_session
-      _expired = create_session expires_at: 1.hour.ago
       _timed_out = create_session timeout_at: 1.hour.ago
 
       assert_equal [valid], Session.valid.all
--- a/app/models/passwordless/session.rb
+++ b/app/models/passwordless/session.rb
@@ -18,8 +18,12 @@ module Passwordless
     before_validation :set_defaults
 
     scope :valid, lambda {
-      where('timeout_at > ? AND expires_at > ?', Time.current, Time.current)
+      where('timeout_at > ?', Time.current)
     }
+
+    def expired?
+      expires_at <= Time.current
+    end
 
     private
 
--- a/test/controllers/passwordless/sessions_controller_test.rb
+++ b/test/controllers/passwordless/sessions_controller_test.rb
@@ -108,5 +108,19 @@ module Passwordless
       assert_equal '/', path
       assert cookies[:user_id].blank?
     end
+
+    test 'trying to sign in with an expired session' do
+      user = User.create email: 'a@a'
+      session = create_session_for user
+      session.update(expires_at: Time.current - 1.day)
+
+      get "/users/sign_in/#{session.token}"
+      follow_redirect!
+
+      assert flash[:error]
+      assert_nil cookies[:user_id]
+      assert_equal 200, status
+      assert_equal '/', path
+    end
   end
 end


mftaff:
So all this would do is block `sign in` if the session is expired. I wasn't sure how to handle the message part. (_Your session has expired, please sign in again._). Do you use flash messaging?

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -96,5 +96,9 @@ def find_session
         token: params[:token]
       )
     end
+
+    def session_expired(session)
```
mikker: I think we should move the responsibility of knowing whether the session is expired to the session itself? Let's move this to a method `Session#expired?`

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -42,7 +42,7 @@ def show
       BCrypt::Password.create(params[:token])
 
       session = find_session
-      sign_in session.authenticatable
+      sign_in session.authenticatable unless session_expired session
```
mikker: I don't think we want to continue this method if the session is expired. We could either _1) do a regular old (boring) `if` statement and return early or _2)_ we could create an `Error` class, `raise` that and `rescue` it at the bottom.

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -42,7 +42,7 @@ def show
...
 
       redirect_enabled = Passwordless.redirect_back_after_sign_in
```
mikker: We don't want to redirect back if the authentication wasn't successful. In that case the user is going to be turned away again and again in an infinite loop.

Instead we would probably want to always redirect to `root_path` and set a `flash[:error]` message about the session being expired.

app/controllers/passwordless/sessions_controller.rb
mftaff: I have come up with a solution using a custom Error. However, rubocop is saying that the show method is too long/complex... Any ideas how to resolve this? (I started messing around with fixing this by squishing the show method down as much as I could, but I thought I'd ask for your input first, before I commit to refactoring the entire show method)

`Metrics/AbcSize: Assignment Branch Condition size for show is too high. [19.1/15]'
'Metrics/MethodLength: Method has too many lines. [16/10]`

```
def show
  # Make it "slow" on purpose to make brute-force attacks more of a hassle
  BCrypt::Password.create(params[:token])
  session = find_session

  begin
     raise ExpiredSessionError if session.expired?
  rescue ExpiredSessionError
     flash[:error] = 'Your session has expired, please sign in again.'
     redirect_to main_app.root_path && return
  end

  sign_in session.authenticatable

  redirect_enabled = Passwordless.redirect_back_after_sign_in
  destination = reset_passwordless_redirect_location!(User)

  if redirect_enabled && destination
     redirect_to destination
  else
     redirect_to main_app.root_path
  end
end
````

Comment (mftaff) [reply app/controllers/passwordless/sessions_controller.rb]: For example here is a refactored view: (rubocop is happy with this)

    def show
      # Make it "slow" on purpose to make brute-force attacks more of a hassle
      BCrypt::Password.create(params[:token])

      session = find_session
      raise ExpiredSessionError if session.expired?

      sign_in session.authenticatable

      redirect_enabled = Passwordless.redirect_back_after_sign_in
      target = reset_passwordless_redirect_location!(User)

      redirect_to(redirect_enabled && target ? target : main_app.root_path)
    rescue ExpiredSessionError
      session_expired
    end
    ...
    private
    ...
    def session_expired
      flash[:error] = 'Your session has expired, please sign in again.'
      redirect_to main_app.root_path
    end

**Comment by mftaff** [comment]
Thanks for the feedback! How does it look now? (I am more than happy to make infinite changes 😄)

File: app/controllers/passwordless/sessions_controller.rb [hunk 4]
```diff
@@ -42,16 +42,16 @@ def show
...
-      destination = reset_passwordless_redirect_location!(User)
+      target = reset_passwordless_redirect_location!(User)
 
-      if redirect_enabled && destination
-        redirect_to destination
-      else
-        redirect_to main_app.root_path
-      end
+      redirect_to(redirect_enabled && target ? target : main_app.root_path)
```
mikker: I like the old way better :)

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -42,16 +42,16 @@ def show
...
+    rescue ExpiredSessionError
+      session_expired
```
mikker: I'd prefer to have the handling logic here instead of its own method.

lib/passwordless/exceptions.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Passwordless
+  # Abstract passwordless errors from StandardError.
+  class PasswordlessError < StandardError; end
+  # Exception raised when a session is expired.
+  class ExpiredSessionError < PasswordlessError; end
```
mikker: While this is not a bad idea, I think I'd prefer to just have the error be a subclass  to the controller. Like:

```ruby
module Passwordless
    class ExpiredSessionError < StandardError; end
    # ...
```

As this is our only error so far and the controller is the only place it'll be raised, I don't think we need to centralize the errors just yet :)

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -96,5 +96,10 @@ def find_session
         token: params[:token]
       )
     end
+
+    def session_expired
+      flash[:error] = 'Your session has expired, please sign in again.'
```
mikker: I think we found our next issue: Adding `I18n` locale support 😄

test/models/passwordless/session_test.rb
```diff
@@ -16,7 +16,7 @@ def create_session(attrs = {})
 
     test 'scope: valid' do
       valid = create_session
-      _expired = create_session expires_at: 1.hour.ago
+      # _expired = create_session expires_at: 1.hour.ago
```
mikker: You can delete this 👍

app/controllers/passwordless/sessions_controller.rb
mftaff: I did too, I did that to try and satisfy rubocop saying the show method was too long... Should I ignore rubocop? or is there a better solution that I missed?

app/controllers/passwordless/sessions_controller.rb
mftaff: 🤣

app/controllers/passwordless/sessions_controller.rb
mikker: Feel free to add ignores to rubocop checks – they're only there for guidance 👍 
The method length one is called `Metrics/MethodLength`

```
# rubocop:disable Metrics/MethodLength
```

and then `# rubocop:enable ...` after the method

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -5,6 +5,8 @@
 module Passwordless
   # Controller for managing Passwordless sessions
   class SessionsController < ApplicationController
+    # Raise this exception when a session is expired.
+    class ExpiredSessionError < StandardError; end
```
mikker: Add a new line under this and then we’re good to go!

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -37,11 +39,13 @@ def create
     #   or _root_path_
     # @see ControllerHelpers#sign_in
     # @see ControllerHelpers#save_passwordless_redirect_location!
-    def show
+    def show # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
mikker: Oh, and move the comment above the method (or above the docs if necessary)

app/controllers/passwordless/sessions_controller.rb
mftaff: Done

app/controllers/passwordless/sessions_controller.rb
mftaff: and done

mikker:
This is good thank you!
💙💚💛💜❤️

mikker:
Oh, I forgot: Can you squash these commits into a single one? Then I'll happily merge it 😄

mftaff:
Uh oh, it looks like I just did something wrong...

mikker:
😀  Not sure but you are right that it looks weird.

Be sure that **a)** your `master` is up to date with my `master` then **b)** `git rebase -i master` this branch on it and `fixup` all your commits into one.

Sorry, can't do much when I don't have it in front of me.

mftaff:
Wow thanks I followed those commands exactly and it did the trick

mikker:
You can even do it on those two `Merge mikker/master` commits :)

mftaff:
Here is the results of running git log -3:

```
commit f364d454427e643305ecbd8ecc939e0cccafc78e (HEAD -> expire-sessions, origin/expire-sessions)
Author: mftaff
Date:   Sun Jan 14 13:37:29 2018 +0000

    Implement expiring sessions.

commit 343352dac833f72bd68c476b76eb18f7c16a3508 (origin/master, origin/HEAD, master)
Merge: 26bbc76 51c1993
Author: mftaff
Date:   Sun Jan 14 12:13:33 2018 +0000

    Merge branch 'master' of https://github.com/mikker/passwordless

commit 51c1993ad4d293ec21938e06f54deb4f78fa142e (tag: v0.4.4, upstream/master)
Author: Mikkel Malmberg
Date:   Tue Jan 2 10:17:08 2018 +0100

    0.4.4
```

When I try running `git rebase -i master` it is only taking me back to that top commit. Is this because the other commits are from my master branch? (When I started today I ran `git pull upstream master` which created that merge commit) Do you know how to resolve this?

(sorry for the technical troubles ;-) hope it isn't too much of a bother)

mikker:
No trouble. Git can be trouble but it's usually because I mess it up myself 😁 

Do you have my fork as a `remote` in your local repo? Then you can reference its `master` like

```
git rebase -i mikker/master # or origin/master, but that could be your own fork
```

If not you can add it with

```
git remote add mikker https://github.com/mikker/passwordless
```

Then the first command should give you the entire history divergence from mine.

IF we end up not getting it to work then it's totally fine. I can remove the two commits with a little force on my end – but it's more fun to see if we can do it _the right way™_.

mftaff:
Yes! it is one commit now 🎆

mikker:
Nice! Ok, _LAST_ thing. I like to keep commit messages in the style of _"Add thing"_ or _"Change things"_. Basically "VERB description of what the app now does differently".

In this case maybe something like _"Redirect user with error when session has expired"_. Makes sense? I read a pretty good blog post on this a while back. Trying if I can find it 😄

mikker:
Still good! https://chris.beams.io/posts/git-commit/

mikker:
_"Imperative mood"_ is the stuff

mftaff:
👍 I will give it a read

mikker:
Nice work! Thank you.

mftaff:
😺 
Again if there is anything else to do I am keen. 
Either way though thanks, I've learned loads from these contributions!!

mikker:
I hope you have! You are very welcome to look into the `I18n` stuff. You don't have to translate anything but rather lay the ground work for users to translate it themselves.

PR #14, opened 2018-03-03
Store routing context when mounting engine, Fix i18n path

grekko:
Disclaimer: This is just a WIP proposal. I am not really sure where/how its best to store the routing configuration for later re-use. Also this needs a test :) Happy to contribute the test.

I noticed that Passwordless allows users to change the root path of the engine. E.g. from the README:

```ruby
passwordless_for :users, at: '/', as: :auth
```

If the root path is changed though, the `Mailer.magic_link` fails to generate the proper `token_sign_in_url`.

The same happens e.g. in the `passwordless/sessions/new.html.erb`-view.

### DIFF
--- a/app/views/passwordless/sessions/new.html.erb
+++ b/app/views/passwordless/sessions/new.html.erb
@@ -1,4 +1,4 @@
-<%= form_for @session, url: send(Passwordless.router_context).sign_in_path do |f| %>
+<%= form_for @session, url: send(Passwordless.mounted_as).sign_in_path do |f| %>
   <% email_field_name = :"passwordless[#{@email_field}]" %>
   <%= text_field_tag email_field_name, params.fetch(email_field_name, nil) %>
   <%= f.submit I18n.t('passwordless.sessions.new.submit') %>
--- a/app/views/passwordless/sessions/new.html.erb
+++ b/app/views/passwordless/sessions/new.html.erb
@@ -1,4 +1,4 @@
-<%= form_for @session, url: send(authenticatable_resource).sign_in_path do |f| %>
+<%= form_for @session, url: send(Passwordless.router_context).sign_in_path do |f| %>
   <% email_field_name = :"passwordless[#{@email_field}]" %>
   <%= text_field_tag email_field_name, params.fetch(email_field_name, nil) %>
   <%= f.submit I18n.t('passwordless.sessions.new.submit') %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -4,7 +4,7 @@ en:
     sessions:
       create:
         session_expired: 'Your session has expired, please sign in again.'
-        email_sent_if_record_found: "If we found you in the system, we have sent you an email."
+        email_sent_if_record_found: "If we found you in the system, we've sent you an email."
       new:
         submit: 'Send magic link'
     mailer:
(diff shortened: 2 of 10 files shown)


mikker:
I actually have a friend stumble into this the other day and didn't have the time to investigate. Perfect! Thanks. I think I'd to have the variable called `mounted_at` instead. Or maybe have a `Passwordless.context` object with a nested `.mounted_at`. This gives us one place to save stuff like this if more shows up.

This is a very nice way to solve this problem 👍 thanks! A test case would be very welcome here as well.

config/locales/en.yml
```diff
@@ -4,7 +4,7 @@ en:
     sessions:
       create:
         session_expired: 'Your session has expired, please sign in again.'
-        email_sent_if_record_found: "If we found you in the system, we've sent you an email."
+        email_sent_if_record_found: "If we found you in the system, we have sent you an email."
```
grekko: I changed wording since rails actionview escapes the rendered content of `we've` into some html entities which were hard to capture in the integration test.

Another solution would be to stub the I18n.translate-call or change the backend/translations just for the test.

Open for discuission here.

test/integration/navigation_test.rb
```diff
@@ -3,7 +3,57 @@
...
+
+    # Submit form
+    post '/users/sign_in',
+      params: {
+        passwordless: { email: alice.email }
+      },
+      headers: { 'HTTP_USER_AGENT' => 'Mosaic v.1' }
+    assert_equal 200, status
+    assert response.body.include?(
+      'If we found you in the system, we have sent you an email.'
```
mikker: `response.body.include?('If we found you in the system')` could be another solution to the `'` business.

test/integration/navigation_test.rb
```diff
@@ -3,7 +3,57 @@
...
+
+    # Follow link, Expect redirect to /secret path which has been unsuccessfully
+    # accessed in the beginning.
+    get token_sign_in_path
+    assert_equal 302, status
+    follow_redirect!
+
+    assert_equal 200, status
+    assert_equal '/secret', path
+    assert_equal 'shhhh! secrets!', response.body
```
mikker: Glorious! 🎉

config/locales/en.yml
mikker: Wondered why you'd changed this. Reason sounds like a real hairball. Fine with this or third option below.

app/views/passwordless/sessions/new.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_for @session, url: send(authenticatable_resource).sign_in_path do |f| %>
+<%= form_for @session, url: send(Passwordless.mounted_as).sign_in_path do |f| %>
```
mikker: Does this mean, we can remove that helper method?

test/integration/navigation_test.rb
grekko: Yea.. I tried swapping the i18n backend/translations or stubbing the calls to it but everything turned out to be extremely cumbersome for this single case.

So eventually I like to go w/ your proposed way of checking for the string partial.

app/views/passwordless/sessions/new.html.erb
grekko: Good catch. Removed it!

mikker:
This is good 👍 
Can I make you consolidate all the commits into one?

grekko:
> Can I make you consolidate all the commits into one?

Sure. But dont you just squash the commits when merging?

mikker:
💙💚💛💜❤️

grekko:
Glad to help 🤗

PR #15, opened 2018-03-03
Fix i18n path for `email_sent_if_record_found`

grekko:
Simply fixes the i18n path.


grekko:
I'd like to add a test for that.

mikker:
Good catch! Thanks – a test would be very welcome!

grekko:
I added an integration test in #14 which also covers and fixes this issue. Therefore I close this one.

PR #37, opened 2019-01-26
Custom expiry time for tokens and sessions

avinoth:


README.md
```diff
@@ -171,6 +172,24 @@ Passwordless.token_generator = -> (session) {
...
 Session is going to keep generating tokens until it finds one that hasn't been used yet. So be sure to use some kind of method where matches are unlikely.
 
+### Token and Session Expiry
+
+#### Token timeout
+
+By default, sign in tokens generated by Passwordless are made invalid after `1 hour` from the time they are generated. If you wish you can override this and supply your custom Proc function that will return a valid datetime object. Make sure the generated time is in the future.
+
+```ruby
+Passwordless.timeout_at = lambda { 2.hours.from_now }
```
mikker: Good idea to make it a proc so we can still use Rails' helpers.

Should we make it even more explicit that people remember to use a proc/lambda? So they don't end up timing out all their sessions two hours after every restart.

README.md
```diff
@@ -171,6 +172,24 @@ Passwordless.token_generator = -> (session) {
 
 Session is going to keep generating tokens until it finds one that hasn't been used yet. So be sure to use some kind of method where matches are unlikely.
 
+### Token and Session Expiry
+
```
mikker: Maybe add a, very short, note on the difference between _timeout and expiration_ here?

README.md
```diff
@@ -171,6 +172,24 @@ Passwordless.token_generator = -> (session) {
...
+#### Token timeout
+
+By default, sign in tokens generated by Passwordless are made invalid after `1 hour` from the time they are generated. If you wish you can override this and supply your custom Proc function that will return a valid datetime object. Make sure the generated time is in the future.
```
mikker: `1.hour`

README.md
```diff
@@ -171,6 +172,24 @@ Passwordless.token_generator = -> (session) {
...
+
+```ruby
+Passwordless.timeout_at = lambda { 2.hours.from_now }
+```
+
+#### Session Expiry
+
+Session expiry is the time when the actual session is itself expired, i.e. users will be logged out and has to sign back in post this expiry time. By default, sessions are valid for `1 year` from the time they are generated. You can override by providing your custom Proc function that returns a datetime object.
```
mikker: `1.year`

README.md
avinoth: How do you mean by explicit? Also, isn't this token timeout, it won't be reset after the restart right? we only call this at the time of session creation.

test/models/passwordless/session_test.rb
```diff
@@ -47,5 +47,25 @@ def call(_session)
 
       Passwordless.token_generator = old_generator
     end
+
+    test 'with a custom expire at function' do
+      custom_expire_at = DateTime.parse('01-01-2100').utc
+      Passwordless.expires_at = lambda { custom_expire_at }
```
mikker: These new settings will leak into the other tests. I think we should either …

1. Save `old_expire = Passwordless.expires_at` up front and then `Passwordless.expires_at = old_expire` at the end.
2. Or make a new `Passwordless::DEFAULTS` hash that we can reset all settings from after every test example.

Pick whatever you want.

test/models/passwordless/session_test.rb
```diff
@@ -47,5 +47,25 @@ def call(_session)
...
+
+      session = Session.new
+      session.validate
+
+      assert_equal custom_expire_at.to_s, session.expires_at.to_s
```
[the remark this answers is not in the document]
avinoth: Both for that and to easily compare the date, otherwise they'll be on different formats and will fail.

mikker:
Perfect! Thanks a bunch for doing this.
💙💚💛💜❤️

PR #42, opened 2019-02-21
Use timeout instead of expired at SessionsController#show

### LINKED ISSUE #38: Session Expires at is wrongly used
Noticed this while doing #37 

We are checking the session expiry time here - https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L48 whereas it should use `timeout_at`. And, the `expires_at` should maybe used in the `authenticate_by_cookie` method instead. Correct me if I'm wrong, I may have misunderstood. If not, let me know, happy to send a PR.

fladson:
Fixes #38 

I didn't get the last part.

> the `expires_at` should maybe used in the `authenticate_by_cookie` method instead

If you can drop a line explaining it more I am happy to help.

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -45,7 +45,7 @@ module Passwordless
       BCrypt::Password.create(params[:token])
 
       session = find_session
-      raise ExpiredSessionError if session.expired?
+      raise ExpiredSessionError if session.timed_out?
 
       sign_in session.authenticatable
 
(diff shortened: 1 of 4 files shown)


fladson:
Not sure why the `timeout_at` is not being updated

mikker:
Hi @fladson! This is great – thanks!

What do you mean by _the timeout_at_ not being updated_?

mikker:
💙💚💛💜❤️

I brought it home in the PR above. Thank you for your contribution!

fladson:
Awesome, thanks for fixing it @mikker!

PR #45, opened 2019-02-22
Convert code to standard style

mikker:
See https://github.com/testdouble/standard.

I've been sceptical about the block rule but want to try it out in a real project before judging so here we are.


mikker:
@searls You happy now?

searls:
yes?

mikker:
Then I’m happy 😊

PR #49, opened 2019-04-02
Pass request to `after-session-save`

aurels:
I would find it useful, what do you think ?

For example : my app is multi-domain and I'd like to know on which domain the user has initiated the login.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ By default, magic link will send by email. You can customize this method. For ex
 config/initializers/passwordless.rb
 
 ```
-Passwordless.after_session_save = lambda do |session|
+Passwordless.after_session_save = lambda do |session, request|
   # Default behavior is
   # Mailer.magic_link(session).deliver_now
 
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -26,7 +26,11 @@ module Passwordless
       session = build_passwordless_session(find_authenticatable)
 
       if session.save
-        Passwordless.after_session_save.call(session, request)
+        if Passwordless.after_session_save.arity == 2
+          Passwordless.after_session_save.call(session, request)
+        else
+          Passwordless.after_session_save.call(session)
+        end
       end
 
       render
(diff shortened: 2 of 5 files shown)


mikker:
Sure. Did you break the tests? 😊

aurels:
Yes I did :p Let me fix them !

aurels:
Not really clear to me how to run them locally.

I did : 

    rbenv shell 2.4.5
    bundle
    cd test/dummy
    bin/rails db:migrate RAILS_ENV=test

and I get : `ActiveRecord::StatementInvalid: SQLite3::SQLException: table "passwordless_sessions" already exists: CREATE TABLE "passwordless_sessions"`

aurels:
Okay, a `bin/rails db:test:prepare` did it ;-)

aurels:
I fixed the specs ;-)

But I realise I introduced a breaking change in the API : `Passwordless.after_session_save` MUST now always have two parameters.

kinduff:
@aurels you could make it optional by assigning `_request = nil` at the lambda. I like this change but I see it's not flexible. How about to optionally be able to pass anything to the `after_session_save` method?

aurels:
@kinduff Yes I can do `_request = nil`.

> How about to optionally be able to pass anything to the after_session_save method?

I thought about that initially but could not find a good way because it's called in `SessionsController`. Any idea ?

mikker:
Perhaps we could use `Proc#arity` to determine how many arguments to pass? https://ruby-doc.org/core-2.2.0/Proc.html#method-i-arity

aurels:
I'll try that !

aurels:
What do you guys think of d4d0428fe81f1ff621592aabd550802dea5ea23b ?

mikker:
LGTM 👍 
Fix the rubocop warnings and I'll be happy to include this.

aurels:
Mmm most of them come from code I didn't write, right ?

mikker:
I’m not even sure at this point. I’ll find out and merge when I find a minute.

mikker:
Merged this in #55 – thanks a bunch!
💙💚💛💜❤️

PR #51, opened 2019-04-25
Add functionality to mark tokens as claimed to prevent re-use

JoeSouthan:
# Description

To help improve the security and one time nature of the links generated by passwordless, I’ve put together a configuration option to expire tokens after a login attempt.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Add authentication to your Rails app without all the icky-ness of passwords.
 * [Usage](#usage)
      * [Getting the current user, restricting access, the usual](#getting-the-current-user-restricting-access-the-usual)
      * [Providing your own templates](#providing-your-own-templates)
+     * [Claming tokens](#claiming-tokens)
      * [Overrides](#overrides)
      * [Registering new users](#registering-new-users)
      * [Generating tokens](#generating-tokens)
@@ -236,6 +237,35 @@ end
 
 You can access user model through authenticatable.
 
+### Claiming tokens
+
+Opt-in for marking tokens as `claimed` so that they can't be used again.
+
+config/initializers/passwordless.rb
+
+```ruby
+# Default is `false`
+Passwordless.claim_token_after_sign_in = true
+```
+
+#### Upgrading an existing Rails app
+
+For small, low traffic apps, you can probably get away with a single migration (not recommended though!), however it best practice to stagger the migrations to minimise locks to the table/database. For a “high availability migration”, follow this [example gist](https://gist.github.com/JoeSouthan/78d3746e8d728d8ab47e67202b59c891).
+
+All in one migration (will lock database during migration, be cautious!):
+
+```bash
+bin/rails generate migration add_claimed_to_passwordless_sessions
+```
+
+```ruby
+class AddClaimedToPasswordlessSessions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :passwordless_sessions, :claimed, :boolean, null: false, default: false
+  end
+end
+
+```
 
 ### Overrides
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -236,6 +237,35 @@ end
 
 You can access user model through authenticatable.
 
+### Claiming tokens
+
+Opt-in for marking tokens as `claimed` so that they can't be used again.
```
mikker: Perhaps _"So they can only be used once"_

README.md
```diff
@@ -236,6 +237,35 @@ end
...
+
+config/initializers/passwordless.rb
+
+```ruby
+# Default is `false`
+Passwordless.claim_token_after_sign_in = true
```
mikker: I think something like `restrict_token_reuse` is more descriptive. What do you think?

README.md
```diff
@@ -236,6 +237,35 @@ end
...
+```
+
+#### Upgrading an existing Rails app
```
mikker: Very nice to include this bit! However I think we should move it into a `<details>` element ([examples](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d)) to avoid distracting people who are not going to use the feature.

We can probably also write a shorter description. Something like ...

> The simplest way to update your sessions table is with a single migration: 

[example]

> Be aware that this will lock your database while migrating. For a _higher availability_ approach, see this [example gist](#).

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -7,6 +7,7 @@ module Passwordless
   class SessionsController < ApplicationController
     # Raise this exception when a session is expired.
     class SessionTimedOutError < StandardError; end
+    class TokenAlreadyUsedError < StandardError; end
```
mikker: Is it _used_ or is it _claimed_? Let's pick one and use that everywhere.

lib/passwordless/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Passwordless
-  VERSION = '0.7.0' # :nodoc:
+  VERSION = "0.7.0" # :nodoc:
```
mikker: Please run `rubocop` and make sure to align.

test/dummy/db/schema.rb
```diff
@@ -19,6 +19,7 @@
     t.text "user_agent", null: false
     t.string "remote_addr", null: false
     t.string "token", null: false
+    t.boolean "claimed", default: false, null: false
```
mikker: For bools like this I like to save them as timestamps. This provides a bit more information for _(as good as)_ free.

Let's change it to `claimed_at:datetime`, default `nil` and provide method in the `Session` class:

```ruby
def claimed?
  !!claimed_at
end
```

JoeSouthan:
Hi @mikker! Thanks for having a look over the PR.

There is the possibility of a bad actor intercepting the token from the email and using it to initiate their own session, the token should immediately be claimed to prevent this. Expiry goes some way to prevent this, but if it's still within the expiry time you can log back in.

README.md
JoeSouthan: Sounds good!

lib/passwordless/version.rb
JoeSouthan: Not sure how this snuck in :smile:

I notice you have `standardrb` shall I use that?

mikker: Yes please. Not sure if hound knows that but we’ll find out. Standard is right and hound should probably be disabled.

JoeSouthan:
@mikker I've removed the section on HA migrations as there are no column constraints/indices being created.

mikker:
💙💚💛💜❤️

PR #53, opened 2019-05-07
Save session.id in session

### LINKED ISSUE #44: Save session.id in cookie instead of user.id to allow expiration
When we `authenticate_by_cookie` we find the user by the saved `user_id` in the cookies. This means we don't actually know when their session expires.

Instead we'll save the session.id and look up the user from that. That way we'll know if the session has expired too. We might even want to check for this on every sign-in-required request?

JoeSouthan:
Fixes #44

I've refactored the sign/in out flow to use the session rather than a cookie.

Couple of things of note:
- This introduces an n+1 query, loading the Passwordless::Session, then the `authenticatable`
  - Could possibly store the Authenticatable in the session as well, but that adds complexity
  - Could employ some Rails caching to prevent loading the Session each time
- This is a breaking change in that `ControllerHelpers.passwordless_for` is now required to set up controllers with the correct resource to authenticate against, however I've made some of the other method params optional. Wondering if it's better to just remove the superfluous params?
- Instead of deprecating `ControllerHelpers#authenticate_by_cookie`, what do you think of removing it since it's a breaking change?

Let me know what you think!

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -41,10 +41,10 @@ module Passwordless
 
       session = passwordless_session
 
-      sign_in session
-
       redirect_enabled = Passwordless.redirect_back_after_sign_in
       destination = reset_passwordless_redirect_location!(User)
+
+      sign_in session
 
       if redirect_enabled && destination
         redirect_to destination
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -39,13 +39,9 @@ module Passwordless
       # Make it "slow" on purpose to make brute-force attacks more of a hassle
       BCrypt::Password.create(params[:token])
 
-      session = find_session
+      session = passwordless_session
 
-      session.claim! if Passwordless.restrict_token_reuse
-
-      raise Passwordless::Errors::SessionTimedOutError if session.timed_out?
-
-      sign_in session.authenticatable
+      sign_in session
 
       redirect_enabled = Passwordless.redirect_back_after_sign_in
       destination = reset_passwordless_redirect_location!(User)
@@ -101,8 +97,8 @@ module Passwordless
       end
     end
 
-    def find_session
-      Session.find_by!(
+    def passwordless_session
+      @passwordless_session ||= Session.find_by!(
         authenticatable_type: authenticatable_classname,
         token: params[:token]
       )
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -5,9 +5,6 @@ require "bcrypt"
 module Passwordless
   # Controller for managing Passwordless sessions
   class SessionsController < ApplicationController
-    # Raise this exception when a session is expired.
-    class SessionTimedOutError < StandardError; end
-
     include ControllerHelpers
 
     # get '/sign_in'
@@ -46,7 +43,7 @@ module Passwordless
 
       session.claim! if Passwordless.restrict_token_reuse
 
-      raise SessionTimedOutError if session.timed_out?
+      raise Passwordless::Errors::SessionTimedOutError if session.timed_out?
 
       sign_in session.authenticatable
 
@@ -58,10 +55,10 @@ module Passwordless
       else
         redirect_to main_app.root_path
       end
-    rescue Session::TokenAlreadyClaimedError
+    rescue Errors::TokenAlreadyClaimedError
       flash[:error] = I18n.t(".passwordless.sessions.create.token_claimed")
       redirect_to main_app.root_path
-    rescue SessionTimedOutError
+    rescue Errors::SessionTimedOutError
       flash[:error] = I18n.t(".passwordless.sessions.create.session_expired")
       redirect_to main_app.root_path
     end
--- a/app/models/passwordless/session.rb
+++ b/app/models/passwordless/session.rb
@@ -4,8 +4,6 @@ module Passwordless
   # The session responsible for holding the connection between the record
   # trying to log in and the unique tokens.
   class Session < ApplicationRecord
-    class TokenAlreadyClaimedError < StandardError; end
-
     belongs_to :authenticatable,
       polymorphic: true, inverse_of: :passwordless_sessions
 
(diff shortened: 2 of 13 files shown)


mikker:
Nice!

- I don't think it's a problem to fetch both Session and Authenticatable. The user can lazy load the Authenticatable and, say, only look for session in their `require_user!` method, then only fetch the associated record if `current_user` is called. I don't think we should add the caching as it'll only add more complexity for a tiny gain.
- I don't really like the new `passwordless_for` in ApplicationController. Could we get the information we need from whatever the user has already put in their routes? I feel like if we have to change the apis it should be towards less _magic_ and more _normal_ Ruby/Rails code.
- What are the benefits of using session over cookie? Can we get the same, long expiry? Then I'm okay with switching. And if we do, you're right that `authenticate_by_cookie` isn't the best name 😁 

Thank you for doing this!

JoeSouthan:
Thanks for taking a look at the PR!

I've
- Refactored it to touch less of the existing API.
- Kept `authenticate_by_session` but removed the controller setup - I agree it was a bit too magic :tada: 

As it stood, cookies did expose implementation details (model type) and allowed for session fixation. I've added `reset_session` as suggested by the [Rails docs](https://guides.rubyonrails.org/security.html#session-fixation-countermeasures) to help mitigate this.

As to expiry, it will now honour whatever is in `Rails.application.config.session_store.expires_at` or `Session#expired?`.

test/controllers/deprecated_secrets_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  def create_session_for(user)
+    Passwordless::Session.create!(
+      authenticatable: user,
+      remote_addr: "yes",
+      user_agent: "James Bond"
+    )
+  end
+
+  def cookie_key(authenticatable_class)
+    DeprecatedSecretsController.new.send(:cookie_key, authenticatable_class)
```
mikker: Let's pull this method out into its own file so we can test it on its own and we don't need weird (in a Rails context) initialisation?

test/controllers/passwordless/sessions_controller_test.rb
```diff
@@ -12,6 +12,10 @@ def create_session_for(user)
       )
     end
 
+    def session_key(authenticatable_class)
+      Passwordless::SessionsController.new.send(:session_key, authenticatable_class)
```
mikker: Same as above.

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -104,8 +97,8 @@ def find_authenticatable
       end
     end
 
-    def find_session
-      Session.find_by!(
+    def passwordless_session
+      @passwordless_session ||= Session.find_by!(
```
mikker: Why do we memoize this if we only use it once?

app/models/passwordless/session.rb
```diff
@@ -33,14 +31,18 @@ def timed_out?
...
-      raise TokenAlreadyClaimedError if claimed?
+      raise Passwordless::Errors::TokenAlreadyClaimedError if claimed?
       touch(:claimed_at)
     end
 
     def claimed?
       !!claimed_at
     end
 
+    def valid_session?
```
mikker: `session.valid_session?` has a double _session_. Let's rename to `session.valid?` ?

lib/passwordless/controller_helpers.rb
```diff
@@ -3,6 +3,12 @@
 module Passwordless
   # Helpers to work with Passwordless sessions from controllers
   module ControllerHelpers
+    # Returns the {Passwordless::Session} (if set) from the session.
+    # @return [Session, nil]
+    def current_passwordless_session(authenticatable_class)
+      @current_passwordless_session ||= Passwordless::Session.find_by(id: session[session_key(authenticatable_class)])
```
mikker: I don't really like to memoize on behalf of the user. My guess is it is only going to lead to weird, hard to figure out inconsistencies? _Rose memoization_ is widely enough known that we can assume people know to do it if they want it.

The readme already suggests to memoize `current_user` which is, probably, mostly what these methods will be used for.

lib/passwordless/controller_helpers.rb
```diff
@@ -27,54 +34,98 @@ def build_passwordless_session(authenticatable)
...
+    # Authenticate a record using the session. Looks for a session key corresponding to
+    # the _authenticatable_class_. If found try to find it in the database.
+    # @param authenticatable_class [ActiveRecord::Base] any Model connected to
+    #   passwordless. (e.g - _User_ or _Admin_).
+    # @return [ActiveRecord::Base|nil] an instance of Model found by id stored
+    #   in cookies.encrypted or nil if nothing is found.
+    # @see ModelHelpers#passwordless_with
+    def authenticate_by_session(authenticatable_class)
+      return unless current_passwordless_session(authenticatable_class)&.valid_session?
+      @current_authenticatable ||= current_passwordless_session(authenticatable_class).authenticatable
```
mikker: Same concern about memoization as above.

lib/passwordless/controller_helpers.rb
```diff
@@ -27,54 +34,98 @@ def build_passwordless_session(authenticatable)
...
-      cookies.encrypted.permanent[key] = {value: authenticatable.id}
-      authenticatable
+    def sign_in(record)
+      passwordless_session = if record.is_a?(Passwordless::Session)
+        record
+      else
+        build_passwordless_session(record).tap { |s| s.save! }
+      end
+
+      sign_out(authenticatable_class)
```
mikker: Is this to make sure to clear the old cookies?

lib/passwordless/controller_helpers.rb
```diff
@@ -27,54 +34,98 @@ def build_passwordless_session(authenticatable)
...
+    # @param (see #authenticate_by_session)
     # @return [boolean] Always true
     def sign_out(authenticatable_class)
+      # Deprecated - cookies
       key = cookie_name(authenticatable_class)
       cookies.encrypted.permanent[key] = {value: nil}
       cookies.delete(key)
+      # /deprecated
+
+      reset_session
```
mikker: Nice find in the docs 👏

mikker:
Thanks a bunch for yet another great addition 💙💚💛💜❤️ !

mikker:
@JoeSouthan I'm not seeing session persistence in my app using this. Sessions behave like session and reset between browser sessions, no matter what I set as `Rails.application.config.session_store.expires_after`.

mikker:
Actually, I think that might've been due to this and not a session vs cookies thing: https://github.com/mikker/passwordless/pull/61

PR #58, opened 2019-08-14
Remove rogue User class in sessions controller

### LINKED ISSUE #57: Support for model that is not named "User"
I did not find the answer of this on the docs (I might have overlooked it).
I am trying to use `passwordless`,  I do not have a model name `User`, the model used in the authentication is named `Teacher`.

So in the routes.rb, I got:
`passwordless_for :teachers`

but when I access to `/teachers/sign_in/my-token` I get:

```
NameError in Passwordless::SessionsController#show
uninitialized constant Passwordless::SessionsController::User
```
Is there a way to use a model NOT named `User`? if so what I am missing?

Note that the view `/teachers/sign_in` works as expected.

Thanks!

mikker:
Fixes #57 and adds a test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Fixes an issue with using a resource class not named `User` ([#58](https://github.com/mikker/passwordless/pull/58))
+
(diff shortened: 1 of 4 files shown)


alexdesi:
@mikker you are faster than Flash! I am going to test your fix soon, thanks a lot!

Issue #21, opened 2018-05-13
Allow customizing destination after sign in


mikker:
We would like to be able customize where users are sent upon successful authentication:

Happens here: https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L58

Should be an option like the others in https://github.com/mikker/passwordless/blob/master/lib/passwordless.rb

(See #19)

dankimio:
It would be nice to have an option to customize flash messages when redirecting the user back.

Could be an option in `save_passwordless_redirect_location` or elsewhere.

mikker:
This was added in https://github.com/mikker/passwordless/pull/69

Issue #19, opened 2018-05-13
specifying redirect url after login


noctivityinc:
How can we change the url of the magic link so we can send a logged in user to a different page then the root url?

mikker:
Hi! This isn't configurable right now. Would love a pull request if you're up for adding it!

Here's where it happens: https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L58
Options are specified in https://github.com/mikker/passwordless/blob/master/lib/passwordless.rb

PR #67, opened 2019-10-10
Feature: default redirect

tiagofsilva:
This PR is about #19 and #21 . But I'm not sure of a couple things:
- If https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L58 is the only place where that custom path is needed (why not if token is already claimed, for ex.?)
- If you liked the config attr name `default_redirect_path`.

Well, would appreciate your guidance on this. 

Thanks!


mikker:
Hi @tiagofsilva! Sorry to leave you hanging for a bit here. Thanks a bunch for doing the initial work! I'm closing in favor of #69. 🙏

Issue #76, opened 2020-02-12
There is should be some global expose on create render


Uysim:
What I want is 

```ruby
- If @resource.present
   = message
- else
   = other_message
```

By override view but is not possible

mikker:
I agree.

At first this was ambiguous on purpose for security reasons but as most apps (I guess) also have some sort of signup form that requires a unique email address, it doesn't provide any security benefits here anyway.

I like your approach of making the resource available and let the app developer choose whether they want to show that one was found only by overriding the view.

PRs welcome 😊

Uysim:
@mikker I submit pull request #77

PR #77, opened 2020-02-14
expose resource to view

Uysim:
global expose on create render 
Example
```erb
<% if @resource.present? %>
  <p>Message</p>
<% else %>
  <p>Otherg</p>
<% end %>
```

Related to: #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,15 @@ app/views/passwordless/sessions/create.html.erb
 app/views/passwordless/mailer/magic_link.text.erb
 ```
 
+Example on create.html.erb
+```erb
+<% if @resource.present? %>
+  <p>Message</p>
+<% else %>
+  <p>Other Message</p>
+<% end %>
+```
+
 See [the bundled views](https://github.com/mikker/passwordless/tree/master/app/views/passwordless).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -122,6 +122,15 @@ app/views/passwordless/sessions/create.html.erb
...
 ```
 
+Example on create.html.erb
+```erb
+<% if @resource.present? %>
+  <p>Message</p>
+<% else %>
+  <p>Other Message</p>
+<% end %>
+```
```
mikker: How about

```
If you'd like to let the user know whether or not a record was found, `@resource` is provided to the view. You may override `app/views/passwordless/session/create.html.erb` for example like so:

```erb
<% if @resource %>
  User found, check your inbox
<% else %>
  No user found with the provided email address?
<% end %>
```

Uysim: I changed it

mikker: I meant to include the first line too instead of “example create.erb”

Uysim: Sorry didn't notice. Not I changed it

mikker:
Wonderful! Thanks for this.

Issue #80, opened 2020-03-29
Store the generated tokens as hashed (BCrypt)


timwis:
Hello, I'd like to use this library but I'm concerned about the fact that the tokens are stored in plain text in the database. For their duration (by default, one hour) they are essentially passwords. I would have expected them to be hashed or at least encrypted before being stored in the database, like a password, no?

I imagine it's less of an issue because they're short-lived, but I guess if an attacker has read access to the database and initiates a login of one of the users, the attacker can get the newly generated token and login as that user. If it were hashed or encrypted, the attacker would need write access to the database and know what hashing/encryption algorithm is used.

I imagine I'm missing something, but passwordless is a pretty close analogue to password reset tokens, and [devise encrypts those](https://github.com/heartcombo/devise/blob/master/lib/devise/models/recoverable.rb#L90).

I'd love to know why that's not the case in this library, because otherwise it looks great!

mikker:
For the same reasons as you state, I didn't find it necessary.

But you are totally right – there are plenty of reasons to hash them. I don't have much time right now during these weird quarantined times, but eventually we ought to hash the tokens. PRs are very welcome!

PR #81, opened 2020-04-05
Store hashed tokens

timwis:
Stores hashed versions of the tokens in the database per #80. This PR is not quite done yet, but I wanted to share the draft to get feedback in the meantime. I'm not normally a ruby developer so I'm sure there's things we could improve in it.

I used `OpenSSL::HMAC` [like devise does](https://github.com/heartcombo/devise/blob/master/lib/devise/token_generator.rb) (and basically copied their token generator class) since BCrypt is not idempotent (there's no way to do `session = Session.find_by(token_hash: BCrypt::Password.create(token))` as `create` generates a new salt every time).

I'll leave more comments inline in the code.

I know I still need to:
- [x] Get the assertion in `navigation_test` to pass
- [ ] Figure out how to use the rails secret key for generating the key name for HMAC, or whether that's even necessary ([devise seems to do it](https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/rails.rb#L42))

Looking forward to your feedback (feel free to edit directly). Hope this is a good start.


app/models/passwordless/session.rb
```diff
@@ -55,10 +57,10 @@ def available?
     def set_defaults
       self.expires_at ||= Passwordless.expires_at.call
       self.timeout_at ||= Passwordless.timeout_at.call
-      self.token ||= loop {
-        token = Passwordless.token_generator.call(self)
-        break token unless Session.find_by(token: token)
-      }
+
+      if not self.token or not self.token_digest
```
timwis, line 61: I couldn't get conditional assignment to work with destructuring 🤷‍♂️

db/migrate/20171104221735_create_passwordless_sessions.rb
```diff
@@ -13,7 +13,7 @@ def change
       t.datetime :claimed_at
       t.text :user_agent, null: false
       t.string :remote_addr, null: false
-      t.string :token, null: false
+      t.string :token_digest, null: false
```
timwis, line 16: Does it make sense to modify the existing migration, or should we add a second migration?

app/models/passwordless/session.rb
```diff
@@ -13,11 +13,13 @@ class Session < ApplicationRecord
       :expires_at,
       :user_agent,
       :remote_addr,
-      :token,
+      :token_digest,
```
timwis, line 16: I thought calling the database column `token_digest` made it more clear that it's not the actual token. The tradeoff is this is a breaking change and assumedly would require a major version bump and running a migration. We can keep the column name the same to make upgrading easier, but you could still call it a breaking change if there are any valid non-hashed tokens in the database when you upgrade.

lib/passwordless.rb
```diff
@@ -4,11 +4,18 @@
...
 require "passwordless/url_safe_base_64_generator"
+require "passwordless/token_generator"
 
 # The main Passwordless module
 module Passwordless
   mattr_accessor(:default_from_address) { "CHANGE_ME@example.com" }
-  mattr_accessor(:token_generator) { UrlSafeBase64Generator.new }
+  mattr_accessor(:token_generator) {
+    Passwordless::TokenGenerator.new(
+      ActiveSupport::CachingKeyGenerator.new(ActiveSupport::KeyGenerator.new('secret')),
```
timwis, line 14: I copied this [from devise](https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/rails.rb#L42). I'm not sure whether it's necessary for `OpenSSL::HMAC` to get a secret "key". If so, should we use rails' `SECRET_KEY_BASE` or whatever?

lib/passwordless/token_generator.rb
```diff
@@ -0,0 +1,23 @@
+require 'openssl'
+
+module Passwordless
+  class TokenGenerator
+    def initialize(key_generator, friendly_token_generator, digest)
+      @key = key_generator.generate_key('passwordless')
```
timwis, line 6: Devise uses the [column name](https://github.com/heartcombo/devise/blob/master/lib/devise/token_generator.rb#L29) but I've just called it `'passwordless'`

test/integration/navigation_test.rb
```diff
@@ -40,13 +40,13 @@ class NavigationTest < ActionDispatch::IntegrationTest
     email = ActionMailer::Base.deliveries.first
     assert_equal alice.email, email.to.first
 
-    # Expect mail body to include session link
-    token_sign_in_path = "/users/sign_in/#{user_session.token}"
-    assert email.body.to_s.include?(token_sign_in_path)
+    # Expect token in email to hash to the one in the session
+    token = email.body.to_s[/\/users\/sign_in\/(.+)/, 1]
+    assert_equal Passwordless.token_generator.digest(token)
```
timwis, line 45: This assertion fails with the error `ArgumentError: wrong number of arguments (given 1, expected 2..3)` and I can't figure out why.

mikker:
Wonderful work so far ❤️

timwis:
Thanks! I’d love your thoughts on the field rename and migration bit when you have a sec (see my inline comments). And the secret key thing if any ideas come to mind, otherwise I’ll do some more research on it.

lib/passwordless.rb
mikker: I think I'd like to keep CachingKeyGenerator out of this block – maybe move it to a `Passwordless.digest(token)` method? Then use it in a setter on `Session` like:

```ruby
def token=(value)
  self.token_digest = Passwordless.digest(value)
end
```

I think using Rails' secret key is a good idea 👍

lib/passwordless/token_generator.rb
```diff
@@ -0,0 +1,23 @@
+require 'openssl'
+
+module Passwordless
+  class TokenGenerator
```
mikker, line 4: I think we should split out the hashing from this. The token generators are supposed to be easy to provide yourself and if you do, I think we should still provide the hashing.

Let's let the token generator still just return a simple, unhashed string.

timwis:
Thanks for the feedback! All great notes. I'll work on that over the next week or so as I get time.

Didn't realise there could be breaking changes in a minor version release for this library though! I'll have to update my project's Gemfile to prevent breaking changes coming in 😬

test/models/passwordless/session_test.rb
```diff
@@ -46,11 +46,46 @@ def create_session(attrs = {})
...
+
+      session = Session.new
+      session.validate
+
+      assert_equal 40, session.token_digest.length
+
+      Passwordless.digest_algorithm = old_digest_algorithm
+    end
+
+    test "with a custom digest secret" do
```
timwis, line 73: This test doesn't make as much sense in the session model test, but I wasn't sure where else to put it, or how to test this feature via the session model.

timwis:
Okay, I think I've incorporated all of your feedback. I dropped the separate TokenGenerator module in order to keep token generation and hashing functionality separate. I moved hashing to a static `Passwordless.digest` method as you suggested, though the only way I could see to do that was on the `Passwordless` module - let me know if there's a better way. I also added a couple tests for the configurable bits.

Beyond this, I think it's just a matter of updating the readme to reflect (1) how to upgrade/migrate, and (2) how to customise hashing.

Looking forward to your feedback.

db/migrate/20171104221735_create_passwordless_sessions.rb
mikker: I think we should modify the original but provide upgrade instructions for existing installations.

app/models/passwordless/session.rb
```diff
@@ -55,10 +57,14 @@ def available?
...
-        break token unless Session.find_by(token: token)
-      }
+
+      if !token || !token_digest
+        self.token, self.token_digest = loop {
+          token = Passwordless.token_generator.call(self)
+          token_digest = Passwordless.digest(token)
+          break [token, token_digest] unless Session.find_by(token_digest: token_digest)
+        }
+      end
```
mikker, line 67: This looks a bit suspicious to me. I don't think anything's wrong but it's hard to make sense of. Do we need/want to check for both `!token || !token_digest`. What if we have one but not the other? Will that ever happen? Could we prevent it? Just dumping my brain, sorry 😊

lib/passwordless.rb
```diff
@@ -9,6 +10,8 @@
 module Passwordless
   mattr_accessor(:default_from_address) { "CHANGE_ME@example.com" }
   mattr_accessor(:token_generator) { UrlSafeBase64Generator.new }
+  mattr_accessor(:digest_algorithm) { "SHA256" }
+  mattr_accessor(:digest_secret) { lambda { Rails.application.secret_key_base } }
```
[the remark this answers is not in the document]
mikker: No need to support anything below 5.1.4 as per https://github.com/mikker/passwordless/blob/master/passwordless.gemspec#L19

lib/passwordless.rb
```diff
@@ -25,4 +28,12 @@ module Passwordless
...
   CookieDeprecation = ActiveSupport::Deprecation.new("0.9", "passwordless")
   SessionValidDeprecation = ActiveSupport::Deprecation.new("0.9", "passwordless")
+
+  def self.digest(token)
+    key_generator = ActiveSupport::CachingKeyGenerator.new(
+      ActiveSupport::KeyGenerator.new(digest_secret.call)
+    )
+    key = key_generator.generate_key("passwordless")
+    OpenSSL::HMAC.hexdigest(digest_algorithm, key, token)
+  end
```
mikker, line 38: Perhaps it would be nice to move this to its own class. Maybe only create the generator once and save it as an ivar.

xdmx:
Any update on this? I'd love to start having hashed values in the db, instead of the real ones.

@timwis were you able to see @mikker's feedback?

timwis:
Hi @xdmx, I haven't been able to integrate @mikker's feedback — unfortunately I've just started a new job that's had me quite full-on :/ I'd love to, but I'd certainly welcome the help if you or @mikker can pick it up! Only a few things left to do, I think.

mikker:
I'm quite busy as well. Any help from anyone on bringing this home is very appreciated 😊

rickychilcott:
It's been a while @timwis  or @xdmx - interested in seeing this across the finish line?

mikker:
AFAIR the work is already done in the big 1.0 PR. It was short sighted of me to have everything be in one PR. If anyone's willing to split it up into separate PRs it would be greatly appreciated.

PR #96, opened 2021-02-17
Add RailsAdmin integration example to readme

### LINKED ISSUE #95: Ideas for compatibility with rails_admin?
Hi @mikker, thanks so much for building Passwordless! It does exactly what I need.

I am using this with [rails_admin](https://github.com/sferik/rails_admin) via the [manual custom auth](https://github.com/sferik/rails_admin/wiki/Customized-authorization) approach.

My user class looks like this:

```rb
#  id         :bigint           not null, primary key
#  email      :string
#  superuser  :boolean          default(FALSE)
#  created_at :datetime         not null
#  updated_at :datetime         not null
#
class User < ApplicationRecord
  passwordless_with :email
  # ...more domain stuff here...
end
```

I've sort of hacked something together that enables access if a user is signed in and has `superuser: true`. It looks like this:

```rb
RailsAdmin.config do |config|
  config.authorize_with do |controller|
    class RailsAdmin::MainController
      include Passwordless::ControllerHelpers
    end
    user = controller.authenticate_by_session(User)
    redirect_to main_app.root_path unless user&.superuser
  end
  # ...more config here...
end
```

However, this doesn't seem ideal. I don't like hacking the main admin controller open every request, but I can't seem to get at the `authenticate_by_session` method any other way.

Do you have any suggestions for what I could try? Happy to PR something to add support for Passwordless into that repo if I can get it working in an ergonomic way.

mplewis:
Closes #95. Thanks for the discussion!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Add authentication to your Rails app without all the icky-ness of passwords.
   * [Redirecting back after sign-in](#redirecting-back-after-sign-in)
   * [Claiming tokens](#claiming-tokens)
   * [Supporting UUID primary keys](#supporting-uuid-primary-keys)
+  * [Integration with RailsAdmin](#integration-with-railsadmin)
 * [Testing helpers](#testing-helpers)
 * [E-mail security](#e-mail-security)
 * [License](#license)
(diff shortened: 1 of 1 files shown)


mplewis:
Hi @mikker, friendly ping on this one :)

mikker:
Thanks! Not going to have time before next week.

mplewis:
Hi @mikker, just wanted to check up on this PR. Anything I can do to help land it?

PR #109, opened 2021-12-17
migrated CI from Travis CI to GitHub Actions workflow, to get CI working again.

madogiwa0124:
This is a PR to migrate to GitHub Action, please close this PR if considering migrating to travis-ci.com, etc.

How about this PR?

## Description

Passwordless uses https://travis-ci.org/, but https://travis-ci.org/ expires on June 15, 2021, so it appears that CI is not currently up and running.

> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.
> https://travis-ci.org/github/mikker/passwordless

In this PR configuration file, we load the repository secret `CODECOV_TOKEN` so that reports are sent to codecov. But since I can't set the token, I checked in test_helper so that I can run CI without specifying it.

https://docs.github.com/en/actions/security-guides/encrypted-secrets

I also excluded ruby2.5, which is EOL, from the CI.

> Ruby 2.5
> status: eol
> release date: 2017-12-25
> EOL date: 2021-03-31
> https://www.ruby-lang.org/en/downloads/branches/

## Test

I have confirmed that the Github Action runs and succeeds in my repository.

https://github.com/madogiwa0124/passwordless/actions/runs/1566667538

PR #110, opened 2021-12-18
Use GH actions for status badge instead of Travis.

madogiwa0124:
This is a PR to changed to status badge to show the result of the GH action.

How about this PR?

## Description

The CI now uses the GH action instead of Travis, but the result badge was still Travis, so I fixed it.

related: https://github.com/mikker/passwordless/pull/109

## Test

I have confirmed that the Github Action result badge appears in my repository.

https://github.com/madogiwa0124/passwordless/blob/migrate-to-gh-action-badge/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <br />
 </p>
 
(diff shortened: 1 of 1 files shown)


rickychilcott:
Thank you @madogiwa0124!

PR #114, opened 2021-12-23
In testing, use the Example domains.

madogiwa0124:
There was a part in the test code where the Example domains was not used, so it has been fixed.

> Example domains
> As described in RFC 2606 and RFC 6761, a number of domains such as `example.com` and `example.org` are maintained for documentation purposes. These domains may be used as illustrative examples in documents without prior coordination with us. They are not available for registration or transfer.
> https://www.iana.org/domains/reserved

How about PR?

### DIFF
--- a/test/controllers/passwordless/sessions_controller_test.rb
+++ b/test/controllers/passwordless/sessions_controller_test.rb
@@ -190,7 +190,7 @@ module Passwordless
     test("signing in and redirecting via insecure query parameter") do
       user = User.create!(email: "a@a")
       passwordless_session = create_session_for(user)
-      get "/users/sign_in/#{passwordless_session.token}?destination_path=http://google.com/secret-alt"
+      get "/users/sign_in/#{passwordless_session.token}?destination_path=http://insecure.example.org/secret-alt"
       follow_redirect!
 
       assert_equal 200, status


mikker:
Thanks! Is there a reason we use the `insecure.` version and not just `https://...`?

mikker (approved):

madogiwa0124:
@mikker 

Thanks for comment!

> Is there a reason we use the insecure. version

This is an insecure redirect destination, we added `insecure.` for clarity.

>  not just https://...

This is because the original implementation was http, so I followed it.

I understand that this test verifies that redirects to another host are properly rejected.

https://github.com/mikker/passwordless/blob/43479f4ea9b80cf646e4f2f8ec0a2cfcaaeda134/app/controllers/passwordless/sessions_controller.rb#L74-L79

As you said, even if don't add a subdomain, it will validate fine because it uses `example.org` which is different from the default `www.example.com` 👍🏻 

Would `https://example.org/secret-alt` or something else be more obvious?

If so, I will try to fix it that way 😃

mikker:
This is fine 👍 I was just curious if it was a conscious choice or not.

mikker:
Thanks! 💙💚💛💜❤️

PR #116, opened 2022-02-07
add controller support for Turbo

kylekeesling:
Hi there mikker - thanks for a great gem! I noticed a small little tweak was needed to support the way Turbo/Hotwire expects responses from controller actions.

I hope this is something you'll find useful and please let me know if you need any other tweaks or additions to this PR.


mikker:
Looks great! Thank you for fixing this. It's a bit unorthodox to render in a `create`, I think that's why I explicitly added the call to render.

💙💚💛💜❤️

kylekeesling:
Happy to help, thanks for a great gem

nickhammond:
@mikker Sounds good, I can take a look next week!

nickhammond:
@mikker After looking at this a bit more and referencing how Slack and Magic.link utilize this flow I think it makes more sense to keep rendering the create.html.erb template instead of utilizing the flash. There's most likely going to be a few additional things that people will want to do that restricting that to flash might be more difficult to work with.

PR #117, opened 2022-02-08
Update README.md

bradgessler:
Simply installation instructions with `bundle add`


bradgessler:
Yes, people execute one command in 1 step instead of 2. I'm working on a PR for bundler that makes this the new default when people create new gems.

PR #120, opened 2022-03-28
Always redirect magic link requests back to the sign_in page and render generic flash

nickhammond:
Hey @mikker,

I realized that I couldn't get magic link requests to work at all without always doing a redirect for Turbo. Turbo is always going to complain about the response needing to be a redirect when it's a simple render call without a `status: :unprocessable_entity`. 

If `sessions/create.html.erb` is removed though I was thinking it might be nice to be able to specify a redirect location instead of the default or some way to detect that a magic link was just requested after the redirect. 

Related to: https://github.com/mikker/passwordless/pull/116
Root issue: https://github.com/hotwired/turbo-rails/issues/12

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -17,7 +17,7 @@ module Passwordless
 
     # post '/sign_in'
     #   Creates a new Session record then sends the magic link
-    #   renders sessions/create.html.erb.
+    #   redirects to sign in page with generic flash message.
     # @see Mailer#magic_link Mailer#magic_link
     def create
       @resource = find_authenticatable
@@ -29,11 +29,10 @@ module Passwordless
         else
           Passwordless.after_session_save.call(session)
         end
-
-        render :create, status: :ok
-      else
-        render :create, status: :unprocessable_entity
       end
+
+      flash[:notice] = I18n.t('passwordless.sessions.create.email_sent_if_record_found')
+      redirect_to(sign_in_path)
     end
 
     # get '/sign_in/:token'
--- a/app/views/passwordless/sessions/create.html.erb
+++ b/app/views/passwordless/sessions/create.html.erb
@@ -1,1 +1,0 @@
-<p><%= I18n.t('passwordless.sessions.create.email_sent_if_record_found') %></p>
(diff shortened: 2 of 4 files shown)


mikker:
Nice debugging, thank you!

I think we'd need a special page with what used to be in `create.html.erb`, as redirecting to sign in is going to confuse people, I think. Something like `completed.html.erb`. Maybe _completed_ is the wrong word. `waiting.html.erb` maybe. What do you think?

nickhammond:
@mikker I have what used to be on the `create.html.erb` page as a flash now so the user still sees it if they have flash configured which is pretty standard. Since the message is a locale(`passwordless.sessions.create.email_sent_if_record_found`) this can easily be overridden at the app level too. 

I suppose not everyone is going to need a full post request magic link flow but right now you can easily override `create.html.erb` which is what I was thinking might need to be catered to. 

What about just another default redirect route?

```
# Default redirection paths
Passwordless.success_redirect_path = '/' # When a user succeeds in logging in.
Passwordless.failure_redirect_path = '/' # When a a login is failed for any reason.
Passwordless.sign_out_redirect_path = '/' # When a user logs out.
Passwordless.session_creation_redirect_path = '/' # When a user requests a magic link
```

sign-in-example

Which makes your setup paths:

1. Do nothing and by default get the redirect to request a magic link with the flash showing
2. Specify a redirect path where the user can expand on whatever additional information they need to explain

vhazbun:
@mikker do you think this will be merged soon?

mikker:
Sure, thank you for this.

mikker:
Apologies, I didn't realize.

I think actually your suggestion to just disable Turbo on our forms is the best and easiest approach. I think the slight downgrade in app speed is acceptable as what the user sees before and after signing is probably not very alike anyway.

ecomba:
I just saw this after bundle upgrading.

As @henrikbjorn commented, it broke my app flow and I decided to revert to the previous version as a flash message will just not do it for me.

henrikbjorn:
@mikker Since we have an easy way of specifying the controller. We could maybe create some helper methods in the `SessionController` so that it is easier overwrite the standard `create` behavior ?

henrikbjorn:
Then just remember to disable turbo on the form, since it requires a redirect...

PR #121, opened 2022-05-09
Smallest PR of the year!

Chocorean:
Just fixing a broken link in the readme... I've been recently blown by a service that uses passwordless auth and I was like, why is this not popular, it's so convenient... Can't wait to use this gem, it looks like you've put a lot of effort into it, thanks for your work! Cheers


mikker:
Thank you @Chocorean!

PR #128, opened 2022-11-08
use form_with, requre email, handle rails 7 form redirect

yshmarov:
Hello @mikker ! 
I've been [playing around](https://github.com/yshmarov/insta2blog.com/pull/75) with adding the gem, and noticed some minor things that could be improved. I hope my suggestions find you well:
* form_with allows more options than form_for, it's easyer to style.
* don't let the html form submit null email
* data-turbo-false added for rails apps that user turbo & no ujs, so that the page does redirect on form submission. otherwise at the moment after form submission there is no redirect
* add email validation regex to the example.
* added ua.yml - one more language for i18n. hopefully native speakers will add other languages in the future. I need this one for my app to work in my region


mikker:
Hey @yshmarov! Thank you for all these. Looks like decent additions.

How does `form_with` benefit styling? I imagine it'll add more or perhaps different ids and classes? Is this potentially a (very subtly and harmlessly) breaking change?

I can definitely see how it's nice to share translations however I fear bundling them in the gem will make it very hard to do updates as we'd have to wait for all translations to be ready before we can do breaking changes, etc... So maybe we can put the translations in the wiki here on GH instead?

It breaks my heart what's being done to your country and people. Rooting for you 💛💙

yshmarov:
Thanks for your reply & support @mikker !

`form_with` is supposed to unify `form_for` and `form_tag`, that seem to be slowly going obsolete https://github.com/rails/rails/pull/26976

> How does form_with benefit styling?

you don't have to apply `id`, `data`, `class` inside html_options
```
<%= form_with model: @post, id: “custom-id”, class: “custom-class” do |form| %>
<%= form_for @post, html: { id: “custom-id”, class: “custom-class” } do |form| %>
```

> Is this potentially a (very subtly and harmlessly) breaking change?

I don't think so...

> I can definitely see how it's nice to share translations however I fear bundling them in the gem will make it very hard to do updates as we'd have to wait for all translations to be ready before we can do breaking changes, etc... So maybe we can put the translations in the wiki here on GH instead?

Good point! You are right!

mikker (changes requested):
- [ ] Move translation to GH wiki
- [x] Add item to changelog

yshmarov:
@mikker I don't see the Wiki tab available on the repo. It might be disabled?

mikker:
Perfect 👌
I think the wiki is there now. I've created the first page.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Changed
+
+- Replaced `form_for` with `form_with`
+- Added frontend validation for email presence in views
```
mikker, line 8:
```suggestion
- Replaced `form_for` with `form_with` in view template ([#128](https://github.com/mikker/passwordless/pull/128))
- Added frontend validation for email presence in views ([#128](https://github.com/mikker/passwordless/pull/128))
```

mikker (approved):
❤️🧡💛💚💙💜

yshmarov:
@mikker looks like the wiki is not community-editable 🙃 
Screenshot 2022-11-30 at 22 04 17

mikker:
Sorry, try again!

Issue #129, opened 2022-11-08
require email presence in SIGN_IN


yshmarov:
if the "send magic link" form is submitted without an email, there will be an SMTP error. 
<img width="851" alt="Screenshot 2022-11-08 at 13 23 35" src="https://user-images.githubusercontent.com/13472945/200562795-e33e6229-eec6-40d9-a6b2-9d284fb9810a.png">

this is because we do not guard against submitting a nil email.

* step 1 - [require email in the HTML form](https://github.com/mikker/passwordless/pull/128/files#diff-a449f0698293d1fd8addd02135de7a25541cba10b3c9bb4674c2743b11643185R3)
* step 2 - [TODO] [render form error](https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L23) if email blank from the controller

fchatterji:
Hey, i'd like to contribute on this issue if possible. Is it still ongoing? I see the first step has already been merged.

mikker:
Please go ahead 😊 PRs welcome

yshmarov:
> 

hey @fchatterji ! 
Having just frontend presence validation is not enough. This way, a user can input an invalid email, successfully submit the form, and get ` Net::SMTPSyntaxError in passwordless/sessions # create `.

I think we can add some validation in [find_authenticable](https://github.com/mikker/passwordless/blob/master/app/controllers/passwordless/sessions_controller.rb#L110) to check if email satisfies `URI::MailTo::EMAIL_REGEXP`...

wilburhimself:
@yshmarov @fchatterji @mikker Created a PR to address this validation in #238

PR #134, opened 2023-02-09
require email on sign in

fchatterji:
Added a check to prevent user submitting an invalid email
github issue: https://github.com/mikker/passwordless/issues/129


mikker:
I don't think this is the way to go. But thank you for contributing this.

PR #143, opened 2023-06-09
pass redirect_to options when signing out

bcasci:
@mikker This is a followup to my earlier PR. 

First, I neglected to notice that the redirect_to options needed to be used on session destroy as well.

Second, in QA testing in my own app, I noticed some unusual behavior. This had to with the Rails `redirect_to` method manipulating the options hash that's passed into it. Since these options are passed by reference, the Passwordless configuration was being changed on the fly by Rails. I got around that by calling `.dup` on those options before passing them to `redirect_to`.


mikker:
Thank you for following up with fixes.

PR #149, opened 2023-06-17
Remove reference to deleted generator file

puffo:
After switching to the redirect strategy in https://github.com/mikker/passwordless/pull/120, there was a leftover reference to the deleted file in the view generator. This causes a non-critical error when running the views generator:

```
$ bin/rails g passwordless:views
      create  app/views/passwordless/mailer/magic_link.text.erb
      create  app/views/passwordless/sessions/new.html.erb
Could not find "sessions/create.html.erb" in any of your source paths. Your current source paths are: ...
```

Here's a quick fix to remove the reference to that file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,13 +118,11 @@ end
 
 Override `passwordless`' bundled views by adding your own. You can manually copy the specific views that you need or copy them to your application with `rails generate passwordless:views`.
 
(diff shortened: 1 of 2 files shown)


mikker:
Good catch! Thank you for contributing a fix

PR #152, opened 2023-06-19
Set a custom controller for `passwordless_for` routes.

henrikbjorn:
Closer to using multiple models in the same project. E.g providing Devise like support for ActiveAdmin but with Passwordless

``` ruby
module Admin
  class SessionsController < Passwordless::SessionsController
    layout 'admin'
  end
end
```

``` ruby
passwordless_for :users
passwordless_for :admin_users, controller: 'admin/sessions'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,20 @@ Please note that, from a security standpoint, this is a **bad practice** because
 
 See [the bundled views](https://github.com/mikker/passwordless/tree/master/app/views/passwordless).
 
+### Using a custom controller
+
+``` ruby
+Rails.application.routes.draw do
+  passwordless_for :users, controller: 'sessions'
+
+  # other routes
+end
+
+class SessionsController < Passwordless::SessionsController
+end
+```
+
+
 ### Registering new users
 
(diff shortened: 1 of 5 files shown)


mikker:
I appreciate the effort!

I feel like if you're already using your own custom controller, you may as well set up your own routes too. It's only 4 lines that are easily copied.

I don't want to have to support an option like this at this point.

henrikbjorn:
> I appreciate the effort!
> 
> I feel like if you're already using your own custom controller, you may as well set up your own routes too. It's only 4 lines that are easily copied.
> 
> I don't want to have to support an option like this at this point.

It seems like a natural extension point instead of copy pasting the helper method, and making sure that the `authenticable` default params etc. are present. 

Previously I defaulted to do something like this:

``` ruby
Rails.configuration.to_prepare do
  Passwordless::SessionsController.class_eval do
    layout 'passwordless'
  end
end
```

Also before I make another PR. I would like to use a custom mailer and thought about adding `Passwordless.mailer` just like `devise` does.

mikker:
Change of heart. This is good! Thank you @henrikbjorn for contributing and for sticking to it.
❤️🧡💛💚💙💜

PR #157, opened 2023-08-11
Use `tableize` instead of `downcase`

henrikbjorn:
In the case of using Passwordless with ActiveAdmin I came across this use. `AdminUser` will be downcased to `adminuser` and then pluralized to `adminusers`. Instead it should be `admin_users`. This is that `tableize` does and in a single method call. So no need to pluralize afterwards


mikker:
Great! Thank you for this

PR #158, opened 2023-08-17
Add a Passwordless::ControllerHelpers#sign_in_authenticatable method

hutchike:
The old version of Passwordless allowed me to sign in a user with a simple `sign_in user` call. The new version requires me to pass a session object instead, so I've added a simple `sign_in_authenticatable user` method to complement it. Hope you like it. I use this when I'm auth'ing via a Google login instead of a sign-in link via email.

### DIFF
--- a/lib/passwordless/controller_helpers.rb
+++ b/lib/passwordless/controller_helpers.rb
@@ -31,7 +31,7 @@ module Passwordless
     end
 
     # Signs in an authenticatable 
-    # @param authenticatable [ActiveRecord::Base] An an instance of a user Model
+    # @param authenticatable [ActiveRecord::Base] an instance of the user Model
     # @return [ActiveRecord::Base|boolean] a Passwordless::Session for the Model
     def sign_in_authenticatable(authenticatable)
       session = build_passwordless_session(authenticatable)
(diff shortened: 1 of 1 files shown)


henrikbjorn:
Would it be better to have a `create_passwordless_session` method, and then have people do `sign_in create_passwordless_session(user)`

I think that is the better API.

mikker:
I see how it would be nice to have a shortcut sometimes, but I think the explicitness of having to go through building the session is almost a feature as you're forced to recognise how it's actually the session that gets "signed in".

What you suggest @henrikbjorn is awfully close to the already existing `build_passwordless_session` method. So close that I don't want to add a new api right now.

hutchike:
Thanks for considering the PR. I just wanted to remind that the original API allowed us to `sign_in` an authenticatable object (e.g. a user), whereas the new version doesn't, so it's a breaking change that will trip many people up. I'm sure there's a more elegant way to restore the ability to sign in an authenticatable object, and I do think this is a feature that's worth retaining. Just my 2c. Thanks again.

mikker:
Thank you for that note. I think you're right. I'll make sure to mention it in the changelog that it's going away.

Issue #169, opened 2023-10-15
enumerate passwordless_sessions


yshmarov:
Currently, after claiming a session it is still possible to navigate to the claimed session URL like http://localhost:3000/sign_in/3.

It is also possible to navigate to something like http://localhost:3000/sign_in/334534543abc and get an error like
```
ActiveRecord::RecordNotFound in Passwordless::SessionsController#show
Couldn't find Passwordless::Session with [WHERE "passwordless_sessions"."id" = $1 AND "passwordless_sessions"."authenticatable_type" = $2]
```

Thus, a user can find how many sessions there were.

I would suggest:
- redirect away from claimed/expoired session paths?
- rescue from RecordNotFound?
- use slugged session ids out of the box?

****

Also I think`sign_in/:id` will be a good candidate for captcha / rate limiting in a production app 🤔

mikker:
Hey! Of the three (four?) what would you suggest as the simplest yet most effective improvement?

yshmarov:
I think the best thing to do is always "use slugged session ids out of the box"
```diff
- /sign_in/:id_slug
- /sign_in/:id
```

Maybe that would mean adding an additional attribute to the session model like `:slugged_id` and having `Session.find_by(slugged_id: params[:id])`

Maybe this could be added as an optional feature... 🤔

yshmarov:
Separately, something like this would prevent being able to view screens for the "old" sessions:
```diff
-@session = Session.find(params[:id])
+@session = Session.not_expired.not_claimed.find(params[:id])
```

mikker:
Could we just use uuids for the slugs? Alternatively: Should the table default to uuids for primary keys instead?

yshmarov:
> Alternatively: Should the table default to uuids for primary keys instead?

Yesss! I think uuids are an even better idea!

mikker:
Thank you for your help @yshmarov! 💙💛

PR #176, opened 2023-10-23
Use UUIDs as indentifiers for sessions in public

mikker:
As per discussion with @yshmarov in #169, a proposal to use UUIDs as Session identifiers in public.

It might be a bit more space and cpu optimized to use UUIDs as the primary keys directly but AFAIK that involves having to turn on a Postgres plugin `pgcrypto` [guides ref](https://guides.rubyonrails.org/v5.0/active_record_postgresql.html#uuid).

I feel like that puts unnecessary hassle on the user's end re: using DBs you don't control, etc. I'm also hesitant to add another required migration upgrade.

This way we can do the switch with only a _vaguely_ breaking change (lol, if that's a thing)

@yshmarov what do you think?

### DIFF
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -18,9 +18,11 @@ ActiveRecord::Schema[7.1].define(version: 2023_06_21_085550) do
     t.datetime "expires_at", precision: nil, null: false
     t.datetime "claimed_at", precision: nil
     t.string "token_digest", null: false
+    t.string "identifier", null: false
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.index ["authenticatable_type", "authenticatable_id"], name: "authenticatable"
+    t.index ["identifier"], name: "index_passwordless_sessions_on_identifier", unique: true
   end
 
   create_table "users", force: :cascade do |t|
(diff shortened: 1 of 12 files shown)


mikker:
@yshmarov Do you have an opinion on this? Also very okay if you don't want to bother, just let me know 😅

yshmarov:
> @yshmarov Do you have an opinion on this? Also very okay if you don't want to bother, just let me know 😅

Heyyy @mikker, I'm loving this PR! 🤗

Having to add postgres plugins as dependencies is not cool very; it's much better if the gem is database-agnostic 👍 

I am happy with adding an additional `identifier` column to the session table 👍

test/dummy/db/schema.rb
```diff
@@ -18,9 +18,11 @@
     t.datetime "expires_at", precision: nil, null: false
     t.datetime "claimed_at", precision: nil
     t.string "token_digest", null: false
+    t.string "identifier", null: false
```
yshmarov, line 21: hmm, with `null: false` it might be worth considering `column "identifier" of relation "passwordless_sessions" contains null values`

yshmarov:
looks lovely
Screenshot 2023-10-31 at 18 58 27
Screenshot 2023-10-31 at 18 58 57

test/dummy/db/schema.rb
mikker: I think we should go with `null: false` as the _new default_ in new installations but `null: true` in the upgrade instructions? That way it'll handle legacy tables and Rails can still `validates :presence`

PR #177, opened 2023-10-24
Update README with the correct command

fermion:
I was following along with the setup instructions and came across this. Hat tip to `rails` for suggesting the correct command.


mikker:
Thanks! ❤️🧡💛💚💙💜

PR #178, opened 2023-11-01
Add default flash notice on SessionsController#destroy

lucianghinda:
## What 

This commit adds a default flash notice to destroy action while keeping the flexibility to customize redirect options via config.

## Why

When creating a session in case of success, a flash notice is displayed, while when destroying a session, there was no flash notice displayed in case of success.

Also, having a successful flash message for destroying a session makes the system tests in the apps using this gem a bit easier to check if the logout happened successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+# Unreleased
+
+### Added
+
+- Add default flash notice for sign out (#178)
+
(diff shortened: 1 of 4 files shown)


mikker:
❤️🧡💛💚💙💜

yshmarov:
heyy @lucianghinda nice to see you here! 🤗

PR #182, opened 2023-11-13
Add paranoid flag to options

divagueame:
Hi, this is my attempt to tackle the paranoid option to hide user existence when requesting token [on Sep 21](https://github.com/mikker/passwordless/issues/131#event-10434257219)
When the paranoid flag is enabled, and a user tries to sign in with a non existing email:
- The same behavior will be shown to the user. 
- No email will be sent to non existing emails/user.
- A session will still be created for non existing emails/user. I am not sure how big of a deal it is, so I left it like that for now.

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -21,20 +21,20 @@ module Passwordless
     #   redirects to sign in page with generic flash message.
     def create
       unless @resource = find_authenticatable
-        raise(
-          ActiveRecord::RecordNotFound,
-          "Couldn't find #{authenticatable_type} with email #{passwordless_session_params[email_field]}"
-        )
+        if Passwordless.config.paranoid
+          @resource = mock_authenticatable
+        else
+          raise(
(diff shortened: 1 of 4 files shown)


app/controllers/passwordless/sessions_controller.rb
```diff
@@ -180,6 +190,13 @@ def find_authenticatable
       end
     end
 
+    def mock_authenticatable
+      email = passwordless_session_params[email_field].downcase.strip
+      resource = authenticatable_class.new(email: email)
+      resource.instance_variable_set(:@mocked, true)
```
mikker, line 196: While this is a clever solution I also think it's a little bit too invasive, I guess. I wonder if we can come up with another solution that completely avoids having to mock anything?

mikker (commented):
Thank you for this. I appreciate how you've done this with the least amount of changes possible. My only concern is the mocked instance var, however I feel it's worth thinking about whether we can do this without it.

mikker:
Thank you for contributing. This feature was implemented in #196 and released in 1.3.0

PR #187, opened 2023-11-28
Allow from address to fall back to parent mailer

dennispaagman:
I'm in a situation where I want the parent mailer class to set the `from` address. 

The use case is that based on the `I18n.locale` we set a different from address/domain.

Right now it's not possible to do this, as even with `nil` for `default_from_address` it will not fall back to the parent (the mailer will fail: ArgumentError (SMTP From address may not be blank)). That basically means only one, fixed from address is supported.

I've come up with a small PR that makes that work by not setting the default value if it's nil.

Curious to hear what you think, I'm not entirely sure about:

* Is the config the right place for that `mailer_defaults` method? Feel like it could also live in the `Passwordless::Mailer` class.
* Does it even need it's own method like this, since there's only one default being set right now. It's easily extendable in the future like this, though.
* ~~I've had some issues with getting all tests to run, not sure if this is correct now. Since we need to change the config just for this test, I _think_ setting it like this and resetting at the end is the best way to do it.~~ edit: found the `WithConfig.with_config` helper that solves my problem 😊
* Probably needs some documentation.

Happy to take it in the direction you want with some pointers 😁

### DIFF
--- a/app/mailers/passwordless/mailer.rb
+++ b/app/mailers/passwordless/mailer.rb
@@ -3,7 +3,7 @@
 module Passwordless
   # The mailer responsible for sending Passwordless' mails.
   class Mailer < Passwordless.config.parent_mailer.constantize
-    default from: Passwordless.config.default_from_address
+    default Passwordless.config.mailer_defaults
 
     # Sends a token and a magic link
     #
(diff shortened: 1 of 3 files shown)


mikker:
Doesn't seem to work. I appreciate your efforts and it looks like a good solution - however, I'm hesitant to add code to solve a problem we already, as you noted, has sort of a solution for.

mikker:
Thanks!

PR #188, opened 2023-11-28
Pass default_url_options from mailer

djfpaagman:
This is a bit related to #187, with the same behavior where we set the hostname dynamically based on locale.

With the introduction of the context in 1.1.0 (I think?) (PR: #180). the url helpers are no longer using the `default_url_options` from the mailer (either via `config.action_mailer.default_url_options` or when overriding them in a mailer themselves via a `default_url_options` method).

This PR fixes that by passing them as arguments to the `url_for` method from the context.

This probably also means you no longer necessarily need to to set a default host for the routes as well, as mentioned here in the docs:

https://github.com/mikker/passwordless/blob/cd0e0ed7cdba745dcae92e84d800122b8d3368b6/README.md?plain=1#L148-L155

I also added a reload when using `WithConfig`, as the parent class for the mailer is set from the config and that can change based on usage of the helper. This probably also needs to be included in #187.

Let me know what you think!

### DIFF
--- a/app/mailers/passwordless/mailer.rb
+++ b/app/mailers/passwordless/mailer.rb
@@ -17,7 +17,8 @@ module Passwordless
         session,
         action: "confirm",
         id: session.to_param,
-        token: @token
+        token: @token,
+        **default_url_options
       )
 
       email_field = session.authenticatable.class.passwordless_email_field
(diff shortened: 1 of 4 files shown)


mikker:
Great solution!

mikker:
The tests are failing. I think maybe due to your refresh mechanism not working as intended?

mikker:
At least it's green! I think it'd be better to move all this from the general `with_config` to the specific test? It is only really relevant to these two examples.

mikker:
Let's move it now and move it back when it becomes necessary. Thanks!

mikker:
Sorry, no, it's good. Thank you!

PR #190, opened 2023-12-05
Skip bcrypt in test env

mikker:
This could make everyone's test suites a bunch faster


mikker:
Some anecdata from my own app:

```
$ time bin/rspec

# before
bin/rspec  59.04s user 4.86s system 88% cpu 1:12.59 total

# after
bin/rspec  28.33s user 4.54s system 74% cpu 44.261 total
```

PR #191, opened 2023-12-08
Update README.md to include CreateUsers migration example

mzrnsh:
In the README, we already give a code sample that creates User model and CreateUsers migration:

```shell
$ bin/rails generate model User email
```

And then we display the code for User model, showing how to mark the passwordless email field:

```ruby
class User < ApplicationRecord
  validates :email,
            presence: true,
            uniqueness: { case_sensitive: false },
            format: { with: URI::MailTo::EMAIL_REGEXP }

  passwordless_with :email # <-- here!
end
```

 Since this snippet already shows the example of how to validate an email field, I believe it's useful to also feature a snippet that adds similar db-level constraint. The idea is to ensure data integrity in case records are modified outside Rails (or within Rails when validations are skipped).

Here's how I do it:

```ruby
class CreateUsers < ActiveRecord::Migration[7.0]
  def change
    create_table :users do |t|
      t.string :email, null: false

      t.index 'LOWER(email)', unique: true, name: 'index_users_on_lowercase_email'

      t.timestamps
    end
  end
end
```

It is not exactly the same as the model validation - this doesn't validate email format. But personally, I don't care because this is only to ensure email uniqueness. If an invalid email is saved forcibly, the worst that's going to happen is that that one user won't be able to log in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,23 @@ class User < ApplicationRecord
 end
 ```
 
+You might want to update the generated migration file to include db-level constraints similar to the model validation above:
+
+```ruby
+class CreateUsers < ActiveRecord::Migration[7.0]
+  def change
+    create_table :users do |t|
+      t.string :email, null: false
+
+      t.index 'LOWER(email)', unique: true, name: 'index_users_on_lowercase_email'
+
+      t.timestamps
+    end
+  end
+end
+```
+
+
 Finally, mount the engine in your routes:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


mzrnsh:
That's a cool new feature. But would it not affect the only model, meaning uniqueness should still be enforced at database level?

mikker:
I feel like we should maybe instead move all the user setup to a separate file, maybe a wiki page. I don't want to add too much to the README as it quickly becomes tl;dr.

On the other hand this is good information and definitely something lesser experienced devs will find helpful and appreciate.

Something like a page called _"Creating a proper User class"_. What do you think?

mzrnsh:
Good idea! Added the wiki page, and updated the README to point to it.

https://github.com/mikker/passwordless/wiki/Creating-the-user-model

mikker:
I like this. Thanks!

PR #194, opened 2024-01-09
Fix issue with `form_with` from Rails

henrikbjorn:
https://github.com/rails/rails/pull/49943

`form_with` cannot have `model: nil` anymore. Make sure to set `@session` if no `authenticable` can be found.


mikker:
Good catch!

PR #195, opened 2024-01-10
update html attributes for secret token form control

### LINKED ISSUE #193: autofocus in the show action view
since the first task in the show view is to enter the secret token, maybe will be a good idea to add autofocus to the input, like in the new action view

https://github.com/mikker/passwordless/blob/c9e5fef80812c6104f94468dfb890292c5ddfa8e/app/views/passwordless/sessions/show.html.erb#L3

https://github.com/mikker/passwordless/blob/c9e5fef80812c6104f94468dfb890292c5ddfa8e/app/views/passwordless/sessions/new.html.erb#L6-L10

luiscobot:
solves #193


mikker:
Thank you!

PR #196, opened 2024-01-19
Add paranoid option

### LINKED ISSUE #131: Add paranoid option to hide user existence when requesting token

yenshirak:
This is a second attempt to implement the paranoid option because the first PR seems stale: #182

closes #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,9 @@ Enable Passwordless on your user model by pointing it to the email field:
 ```ruby
 class User < ApplicationRecord
   # your other code..
-  
+
   passwordless_with :email # <-- here! this needs to be a column in `users` table
-  
+
   # more of your code..
 end
 ```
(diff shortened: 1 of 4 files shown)


mikker:
I like this implementation a lot better. Thank you for contributing!

Would you mind adding changelog entries for your recent additions? I'll push a minor release then 🌈

yenshirak:
Sure, I'm on it.

PR #197, opened 2024-01-19
Add only_path param to system test helpers

yenshirak:
With `only_path` set to `true` there's no need to set the host.


mikker:
Thank you!

PR #198, opened 2024-01-19
Use default from address of parent when default_from_address is nil

yenshirak:
I think it makes sense to use the default from address of the parent to avoid duplication. I think this isn't a breaking change because it made no sense to set the option `nil` before.


mikker:
Thank you!

PR #199, opened 2024-01-19
Add parent controller option

yenshirak:
I think this makes sense because it follows the same pattern as `parent_mailer`. It is possible to extract the base controller logic into a concern and include that in a custom Passwordless controller but using an option would be more convenient.


mikker:
Do you have a need for this in your app(s)?

yenshirak:
Yes. It's used on the admin interface so the parent should be the admin base controller.

mikker:
Got it. Thanks!

PR #201, opened 2024-01-24
Update README.md

luiscobot:


mikker:
Is this not already configured in Rails? I've never seen this before

mikker:
Thank you!

Issue #206, opened 2024-02-01
Support scoped routes


mikker:
Currently Passwordless' routes need to be defined at the root level. Would be nice if we could support scoping and nesting, like requested in #205, ie:

```ruby
scope ":locale" do
    passwordless_for :users
```

PR #208, opened 2024-02-13
locale to support scoped routes #206

andreaskundig:
This seems to fix the problem. 

The tests pass, however I don't know how to set up a test for a scoped locale. I'm new to rails, so it's better to check this very closely.

### DIFF
--- a/lib/passwordless/config.rb
+++ b/lib/passwordless/config.rb
@@ -44,8 +44,10 @@ module Passwordless
     option :sign_out_redirect_path, default: "/"
     option(
       :after_session_save,
-      default: lambda do |session, _request|
-        Mailer.sign_in(session, session.token).deliver_now
+      default: lambda do |session, request|
+        locale = request.params[:locale]
+        url_options = locale ? { locale: locale } : {}
+        Mailer.sign_in(session, session.token, url_options).deliver_now
       end
     )
 
--- a/lib/passwordless/config.rb
+++ b/lib/passwordless/config.rb
@@ -44,10 +44,8 @@ module Passwordless
     option :sign_out_redirect_path, default: "/"
     option(
       :after_session_save,
-      default: lambda do |session, request|
-        locale = request.params[:locale]
-        url_options = locale ? { locale: locale } : {}
-        Mailer.sign_in(session, session.token, url_options).deliver_now
+      default: lambda do |session, _request|
+        Mailer.sign_in(session, session.token).deliver_now
       end
     )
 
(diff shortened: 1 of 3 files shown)


lib/passwordless/config.rb
```diff
@@ -44,8 +44,10 @@ class Configuration
     option :sign_out_redirect_path, default: "/"
     option(
       :after_session_save,
-      default: lambda do |session, _request|
-        Mailer.sign_in(session, session.token).deliver_now
+      default: lambda do |session, request|
+        locale = request.params[:locale]
```
mikker, line 48: I don't think `params[:locale]` is a built-in Rails thing? I've seen it before but is it not just a typical thing rather than an officially supported thing?

mikker (commented):
I like this, just a comment

lib/passwordless/config.rb
andreaskundig: It's not built-in but it's all over the official [i18n rails guide](https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests). This is for example the only documentation I find about how to use the built-in *I18n.with_locale*:

``` ruby
around_action :switch_locale

def switch_locale(&action)
  locale = params[:locale] || I18n.default_locale
  I18n.with_locale(locale, &action)
end

```

mikker: I don't like to include the `params[:locale]` part by default. I definitely like there to be a way to support it, so the `url_options` param is good. Is there a way we can support something like `params[:locale]` (or whatever) without putting it in there?

I think it'll have to be like that.

andreaskundig: I reverted the changes involving `params[:locale]` (in sessions_controller.rb and context.rb). 

For my use case it means I can use the passwordless Mailer with the new url_options, but the only way I can get things to work in my project is to implement my own equivalent of `session_controller#create` and `after_session_save`.

andreaskundig: Another solution would be to add an optional configurable `make_url_options(request)` function somewhere. If the user implements it, it would be used in the `session_controller#create` and `after_session_save`. That avoids introducing `:locale` in the code, but adds an option that's difficult to explain...

andreaskundig: Ok so I just found out how to get it working without touching the mailer. Using the default_url_options is enough!

`config/initializers/passwordless.rb`
```
# passwordless.rb
  config.parent_mailer = "PasswordlessMailer"
end
```
`app/mailers/passwordless_mailer.rb`
``` ruby
class PasswordlessMailer < ActionMailer::Base

  # adds params[:locale] to every url_for, link_to etc...
  def default_url_options
    Rails.application.config.action_mailer.default_url_options.merge(
      locale: I18n.locale
    )
  end
end
```

This should work with the session_controller too, except that `Passwordless.context.path_for` fails to use the default_url_options! I fix that in pull request [#209](https://github.com/mikker/passwordless/pull/209)

mikker:
Thank you for working on this!

PR #215, opened 2024-04-06
replace flash[:error] with flash.alert

luiscobot:
Since Rails does not include a `:error` flash type, I think we should use the `flash.alert` flash type that [Rails provides](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash.html).


mikker:
Aboslutely! Thank you ❤️🧡💛💚💙💜

PR #218, opened 2024-04-24
Disable autosave of session.authenticatable

### LINKED ISSUE #217: Paranoid mode still autosaves its associated authenticatable model
When `config.paranoid = true`, the authenticatable model is automatically created and saved to the database.  This is unexpected and unwanted.

The `authenticatable` relationship is defined here:
https://github.com/mikker/passwordless/blob/971093ee814df13ec0b485840cb8fbd55d2179db/app/models/passwordless/session.rb#L9-L13

From https://api.rubyonrails.org/v7.1.3/classes/ActiveRecord/AutosaveAssociation.html
> Note that `autosave: false` is not same as not declaring `:autosave`. When the `:autosave` option is not present then new association records are saved but the updated association records are not saved.

So, when `handle_resource_not_found` does this:
https://github.com/mikker/passwordless/blob/971093ee814df13ec0b485840cb8fbd55d2179db/app/controllers/passwordless/sessions_controller.rb#L203-L206
And then does this:
https://github.com/mikker/passwordless/blob/971093ee814df13ec0b485840cb8fbd55d2179db/app/controllers/passwordless/sessions_controller.rb#L23-L26

...assuming there are no failing validations on `@resource`, then both the Passwordless::Session _and_ the authenticatable association are saved.

We worked around this by adding the following to our `config/initializers/passwordless.rb`:
```ruby
reflection = Passwordless::Session.reflect_on_association :authenticatable
reflection.options[:autosave] = false
```

As a simple fix, I propose adding `autosave: false` to the association.  Although, that may result in breaking the "paranoid" aspect of `config.paranoid = true`.  I haven't looked into the best way to handle this in the controller (yet), and automatically creating the authenticatable record was the more immediately pressing issue for us.

nevans:
This prevents the authenticable model from being auto-created when `config.paranoid = true`.

Fixes #217.

### DIFF
--- a/app/models/passwordless/session.rb
+++ b/app/models/passwordless/session.rb
@@ -9,7 +9,8 @@ module Passwordless
     belongs_to(
       :authenticatable,
       polymorphic: true,
-      inverse_of: :passwordless_sessions
+      inverse_of: :passwordless_sessions,
+      autosave: false
     )
 
     validates(


mikker:
Included in #219. Thank you so much for diagnosing, reporting and fixing this!

PR #221, opened 2024-04-25
Drop explicit I18n load path configuration

Dakota-Schramm:
There is currently a problem with locales load order such that in production, passwordless locale files come after the application locale files in the I18n load path

This prevents the application from overriding any translations

Rails engines don't need to explicitly configure this at least since 5.1.4 (the current minimum requirement)
  - https://api.rubyonrails.org/v5.1.4/classes/Rails/Engine.html


Dakota-Schramm:
We tested this in our own production specifically for the subject line and found that this change works

mikker:
Great, thank you!

Must be a leftover from way back when.

PR #224, opened 2024-04-27
pass token to email subject translation

luiscobot:
I think it is better to have the option to pass the token in the email subject, so we do not need to even open the email, maybe we can only see the notification in the phone and right there will be the token.

With this change we could use the token in the email subject locales key.

CleanShot 2024-04-26 at 19 14 22@2x


mikker:
Thank you for contributing this.

Does it complain if you don't use the token, like we don't in the default?

PR #225, opened 2024-05-03
Unknown email sender config

Dakota-Schramm:
Currently, there is no configuration option to allow for sending an email when the resource is not found when the user opts into the "paranoid" option.

With this addition, users will have a special email sent to them when their email is not found in the database


mikker:
Thank you for contributing this.

Is this mainly a security measure (someone's trying to be you) or user experience (you don't have an account, try another address)?

mikker:
My feeling is that this functionality shouldn't be included. Feel free to reopen if you'd like to argue otherwise 😊

Thank you for the suggestion and for using Passwordless though!

PR #226, opened 2024-05-09
Update README.md

ericgross:
Update the readme to indicate that `restrict_token_reuse` defaults to `true` rather than `false`.


mikker:
Thank you for catching and contributing this 👍

Issue #139, opened 2023-05-25
Feature Request: Constraint helpers for `config/routes.rb` 


henrikbjorn:
When using devise they have helpers that can be used in `config/routes.rb` like:

``` ruby
authenticated :admin do
  root to: 'admin/dashboard#show', as: :admin_root
end
```

Which i personally like very much and better than doing `require_user!` in my controllers.

Would be a awesome addition to `passwordless`.

Not sure how it would work with Passwordless as most the user finding and authorizing etc. is in the ControllerHelper where Devise calls Warden e.g: https://github.com/heartcombo/devise/blob/main/lib/devise/rails/routes.rb#L477

mikker:
Good question! PRs welcome ❤️

yshmarov:
on a custom app that I built, something like this works for me:

```ruby
# app/constraints/user_constraint.rb
class UserConstraint
  def initialize(&block)
    @block = block
  end

  def matches?(request)
    user = current_user(request)
    user.present? && @block.call(user)
  end

  def current_user(request)
    User.find_by(id: request.session[:user_id])
  end
end
```

```ruby
# config/routes.rb
  # authenticate :user, ->(user) { user.admin? } do # <- devise syntax
  constraints UserConstraint.new { |user| user.admin? } do
    mount GoodJob::Engine, at: "good_job"
    mount Avo::Engine, at: Avo.configuration.root_path
  end

  # authenticated :user do # <- devise syntax
  constraints UserConstraint.new { |user| user.present? } do
    get 'dashboard', to: 'static#dashboard'
  end
```

tdegrunt:
Similarly to the above, but the following works for me with Passwordless:

```ruby
# app/constraints/user_constraint.rb
class PasswordlessConstraint
  include Passwordless::ControllerHelpers

  attr_reader :authenticatable_type, :session, :lambda

  def initialize(authenticatable_type, lambda)
    @authenticatable_type = authenticatable_type
    @lambda = lambda
  end

  def matches?(request)
    @session = request.session
    authenticatable = authenticate_by_session(authenticatable_type)
    authenticatable.present? && lambda.call(authenticatable)
  end
end
```

```ruby
# config/routes.rb
constraints PasswordlessConstraint.new(User, ->(user) { user.has_role?("admin") }) do
  root to: 'admin/dashboard#show', as: :admin_root
end
```

The rest is syntactic sugar, but a bit too much magic.

mikker:
[...]

PR #227, opened 2024-05-23
Constraint for config/routes.rb

tdegrunt:
Should fix https://github.com/mikker/passwordless/issues/139

- Renames the constraint to Passwordless::Constraint
- Added tests
- Added docs


henrikbjorn:
Could the `lambda` be made optional ?

tdegrunt:
Updated it to support both suggestions. I had thought about making the proc optional, but I would've been okay with just using `-> { true }`, but this is nicer.

mikker:
Included in #228. Thank you for contributing to Passwordless!

PR #229, opened 2024-08-01
Update documentation to reflect implementation

dewey:
Ran into that when copying from documentation.


mikker:
Ah, sorry. Good catch, thanks!

PR #231, opened 2024-09-05
Prompt to change belongs_to type for e.g. UUIDs

StevenClontz:
Took me the longest time to figure out why I couldn't authenticate: I've set up my models to use UUIDs, and it takes a manual step to set up the migration to be compatible with this. This will prompt the next developer to avoid the two hours I lost figuring this out. (-:


mikker:
I like being explicit to help others find it 👍
Thanks!

mikker:
Test failure is only `rails#master`

PR #237, opened 2024-10-07
Add after_session_confirm hook after session confirmation

### LINKED ISSUE #213: Add `after_session_confirm` hook after session confirmation
Just a suggestion to add a `after_session_confirm` similar to `after_session_save`. This would let folks mark accounts as "verified" or "has_verified_email" after session confirmation.

wilburhimself:
fixes #213 

This PR introduces a new `after_session_confirm` hook to the Passwordless gem. This feature allows users to perform custom actions after a session has been successfully confirmed, such as marking an email as verified or updating user attributes after their first successful login.

## Changes
- Added `after_session_confirm` configuration option
- Updated `SessionsController` to call the hook after successful authentication
- Added new tests in `SessionsControllerTest` to cover the new functionality
- Updated README with documentation for the new feature

## Motivation
Users of the gem often need to perform specific actions after a successful login, such as updating user attributes or triggering other processes. The `after_session_confirm` hook provides a clean, configurable way to achieve this without modifying the gem's core code.

## Implementation Details
- The hook is called in the `authenticate_and_sign_in` method of `SessionsController`, right after successful authentication but before redirecting the user.
- The hook can be configured to accept either one argument (the session) or two arguments (the session and the request), providing flexibility for different use cases.
- If the hook is not defined or doesn't respond to `call`, it is silently ignored to maintain backwards compatibility.

## How to Use
Users can configure the hook in their Passwordless configuration:

```ruby
  config.after_session_confirm = ->(session) {
    user = session.authenticatable
    user.update(email_verified: true)
  }
end
```

## Testing
New tests have been added to `SessionsControllerTest` to ensure the hook is called correctly in various scenarios:
- Successful authentication
- Hook receiving both session and request objects
- Hook not being called on failed authentication

All existing tests pass with these new changes.

## Reviewer Notes
- Please pay special attention to the placement of the hook call in `SessionsController` to ensure it's in the most appropriate location.
- Consider if any additional error handling is needed around the hook execution.
- Let me know if you think any additional tests or documentation would be beneficial.

## Checklist
- [x] Code follows the style guidelines of the project
- [x] New tests added for the feature
- [x] All tests pass
- [x] Documentation (README) has been updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ Passwordless.configure do |config|
 
   config.paranoid = false # Display email sent notice even when the resource is not found.
 
-  config.after_session_confirm = ->(session) {} # Called after a session is confirmed.
+  config.after_session_confirm = ->(request, session) {} # Called after a session is confirmed.
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -266,24 +266,20 @@ Passwordless.configure do |config|
 end
 ```
 
-## After Session Confirmation Hook
-The after_session_confirm hook is called automatically after a successful session confirmation. It receives the request and session objects as arguments, which you can use to access the authenticated user (via session.authenticatable) and perform any necessary operations.
+## After Session Confirm Hook
 
-You can configure the after_session_confirm hook in your Passwordless configuration:
+An `after_session_confirm` hook is called after a successful session confirmation – in other words: after a user signs in successfully.
 
 ```ruby
 Passwordless.configure do |config|
-  # ... other configuration options ...
-
   config.after_session_confirm = ->(session, request) {
-    # Your custom logic here
     user = session.authenticatable
-    user.update(email_verified: true)
-    user.update(last_login_ip: request.remote_ip)
+    user.update!(
+      email_verified: true.
+      last_login_ip: request.remote_ip
+    )
   }
 end
-```
-
 ### Token generation
 
 By default Passwordless generates short, 6-digit, alpha numeric tokens. You can change the generator using `Passwordless.config.token_generator` to something else that responds to `call(session)` eg.:
--- a/README.md
+++ b/README.md
@@ -248,49 +248,6 @@ Passwordless.configure do |config|
 end
 ```
 
-## After Session Confirmation Hook
-Passwordless now supports an after_session_confirm hook that allows you to perform custom actions after a session has been successfully confirmed. This is particularly useful for tasks such as marking an email as verified or updating user attributes after their first successful login.
-
-#### Configuration
-You can configure the after_session_confirm hook in your Passwordless configuration:
-
-```ruby
-Passwordless.configure do |config|
-  # ... other configuration options ...
-
-  config.after_session_confirm = ->(session) {
-    # Your custom logic here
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -243,6 +243,51 @@ Passwordless.configure do |config|
   config.sign_out_redirect_path = '/' # After a user signs out
 
   config.paranoid = false # Display email sent notice even when the resource is not found.
+
+  config.after_session_confirm = ->(session) {} # Called after a session is confirmed.
```
mikker, line 247:
```suggestion
  config.after_session_confirm = ->(request, session) {} # Called after a session is confirmed.
```

Let's just always send req + session in this. The other one is mostly there to keep the api backwards compatible.

README.md
```diff
@@ -264,6 +266,24 @@ Passwordless.configure do |config|
...
+
+  config.after_session_confirm = ->(session, request) {
+    # Your custom logic here
+    user = session.authenticatable
+    user.update(email_verified: true)
+    user.update(last_login_ip: request.remote_ip)
+  }
+end
+```
+
```
mikker, line 286:
```suggestion
## After Session Confirm Hook

An `after_session_confirm` hook is called after a successful session confirmation – in other words: after a user signs in successfully.

```ruby
  config.after_session_confirm = ->(session, request) {
    user = session.authenticatable
    user.update!(
      email_verified: true.
      last_login_ip: request.remote_ip
    )
  }
end
```

```

File: lib/passwordless/config.rb [hunk 3]
```diff
@@ -51,6 +51,12 @@ class Configuration
 
     option :paranoid, default: false
 
+    option(
+      :after_session_confirm,
+      default: lambda do |session, _request|
```
mikker, line 56:
```suggestion
      default: lambda do |_session, _request|
```

mikker:
Failing test is just `rails#main`

mikker:
Great work! ❤️🧡💛💚💙💜

wkirby:
@mikker when can we expect to see this released?

mikker:
Just pushed 1.8.0 with this change 😊

henrikbjorn:
Am I correct in assuming the controller context is lot in the hook ?

A use case would be for Ahoy to call `ahoy.authenticate` for analytics. But it seems that can't be done since the hook is "just" a function.

Would it make sense to instead define the hook methods on the controller (kind of like Devise)? Then the Controller context is kept, and it is also easy to set a custom controller to use.

henrikbjorn:
Another note. It is not possible to have two different hooks. E.g one for AdminUsers and one for normal Users. Which having it as a controller method would handle.

mikker:
Good points Henrik!

It could be useful to have the controller context at hand.

Maybe something like if setting the hook setting to a `:symbol`, a method with that name will be called instead?

Like

```ruby
class ApplicationController < ...
  private

  def pwless_after_session_confirm(session) # no need to pass request either
    # ...
  end
end

  config.after_session_confirm = :pwless_after_session_confirm
end
```

That would make it a non-breaking change.

PR #239, opened 2024-12-12
Custom redirect locations

StevenClontz:
This PR extends the behavior of `save_passwordless_redirect_location!` to support saving custom URLs, and uses it to support implementing `redirect_to` parameter on the `#new` action. Example usage:

```ruby
<%= link_to "Sign up or sign in", users_sign_in_path(redirect_to: request.url) %>
```

### DIFF
--- a/lib/passwordless/controller_helpers.rb
+++ b/lib/passwordless/controller_helpers.rb
@@ -87,9 +87,11 @@ module Passwordless
     # Saves request.original_url as the redirect location for a
     # passwordless Model.
     # @param (see #authenticate_by_session)
+    # @param url [String] the url to redirect to
     # @return [String] the redirect url that was just saved.
-    def save_passwordless_redirect_location!(authenticatable_class)
-      session[redirect_session_key(authenticatable_class)] = request.original_url
+    def save_passwordless_redirect_location!(authenticatable_class, url=nil)
+      url = request.original_url if url.nil?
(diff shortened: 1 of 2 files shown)


lib/passwordless/controller_helpers.rb
```diff
@@ -87,9 +87,11 @@ def sign_out(authenticatable_class)
     # Saves request.original_url as the redirect location for a
     # passwordless Model.
     # @param (see #authenticate_by_session)
+    # @param url [String] the url to redirect to
     # @return [String] the redirect url that was just saved.
-    def save_passwordless_redirect_location!(authenticatable_class)
-      session[redirect_session_key(authenticatable_class)] = request.original_url
+    def save_passwordless_redirect_location!(authenticatable_class, url=nil)
+      url = request.original_url if url.nil?
+      set_passwordless_redirect_location! url
```
StevenClontz, line 94:
```suggestion
      session[redirect_session_key(authenticatable_class)] = url
```

mikker:
Interesting! Not saying this is bad – would `config.success_redirect_path` suffice? Or do you specifically want to set it to different things depending on context/page or something like that?

StevenClontz:
> Or do you specifically want to set it to different things depending on context/page or something like that?

Exactly. Feedback on my app was that users would click my sign in button, enter their code, and then have trouble navigating back to the page they were at when they originally tried to log in.

Issue #243, opened 2025-01-22
Clean up expired sessions


duarme:
Currently, expired sessions sit indefinitely on the database.

I think we should at least add to the README a reminder to call `Passwordless::Session.where('expires_at < ?', Time.zone.now).delete_all` in a task that runs periodically.

What do you think?

PR #244, opened 2025-01-22
fixes issue #243

duarme:
Adds a reminder to clean up expired sessions into README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,10 @@ Rails.application.config.session_store :cookie_store,
   # ...
 ```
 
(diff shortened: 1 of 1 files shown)


duarme:
I've been using passwordless for years and realized I had the database full of expired sessions only a few days ago. A heads-up in the README would have been helpful in my case. Anyway, no problem, I still like this, **a lot**.

mikker:
It's not really a problem unless someone feels it's a problem for them and when they decide it is, it's relatively easy to clean up manually.

Thank you!

PR #247, opened 2025-02-11
Fix Rails 8 route lazy loading issue

### LINKED ISSUE #246: `No resource registered` error in the development and test environments with Rails 8
`Passwordless::Context#resource_for` assumes that the routes are already loaded because `Passwordless::Context#resources`is populated by `Passwordless::RouterHelpers::passwordless_for`:
```
def resource_for(session_or_authenticatable)
  if session_or_authenticatable.is_a?(Session)
    session_or_authenticatable = session_or_authenticatable.authenticatable.model_name.to_s.tableize.to_sym
  end

  resources[session_or_authenticatable.to_sym]
end
```

The problem is that `config.eager_load = false` in the development environment and if the `Passwordless::Mailer#sign_in` email is sent asynchronously from a background job in Rails 8 then `resources` may be empty which will cause a `No resource registered` error.

There is another problem in the test environment as well when you call `passwordless_sign_in` before visiting a path.

Calling `Rails.application.reload_routes_unless_loaded` before calling `resource_for` would solve the issue. The question is where to call it.

As workaround I added a SolidQueue `on_start` hook to fix the development environment issue:
```
if Rails.env.development?
  SolidQueue.on_start do
    Rails.application.reload_routes_unless_loaded
  end
end
  ```

And an RSpec `before` hook to fix the test environment issue:
```
config.before(:suite) do
  Rails.application.reload_routes_unless_loaded
end
```

Devise had a similar problem: https://github.com/heartcombo/devise/pull/5728
Rails deferred route loading PR: https://github.com/rails/rails/pull/52353

I opened a PR with a possible fix.

What do you think?

yenshirak:
closes #246

### DIFF
--- a/lib/passwordless.rb
+++ b/lib/passwordless.rb
@@ -17,6 +17,9 @@ module Passwordless
   def self.context
     return @context if @context
 
+    # Routes are lazy loaded in Rails 8 so we need to load them to populate Context#resources.
+    Rails.application.try(:reload_routes_unless_loaded)
+
     LOCK.synchronize do
       @context ||= Context.new
     end


timwis:
Thanks for the fix! @mikker, would you mind publishing to rubygems when you have a moment?

PR #249, opened 2025-02-24
Case insensitive tokens

### LINKED ISSUE #248: Support case-insensitive tokens?
My token generator uses only uppercase letters and numbers. When hand-entering the token, sometimes the user is lazy and writes letters in lowercase, which I want to allow. I'm supporting this by re-implementing `Passwordless::Session`, but it'd be nice if this was a configuration option.

StevenClontz:
closes #248

### DIFF
--- a/app/models/passwordless/session.rb
+++ b/app/models/passwordless/session.rb
@@ -32,9 +32,10 @@ module Passwordless
     # hashed version in the database
     attr_reader :token
 
-    def token=(plaintext)
-      self.token_digest = Passwordless.digest(plaintext)
-      @token = (plaintext)
+    def token=(token)
+      token = token.upcase if Passwordless.config.case_insensitive_tokens
+      self.token_digest = Passwordless.digest(token)
+      @token = token
     end
 
     def authenticate(token)
--- a/test/models/passwordless/session_test.rb
+++ b/test/models/passwordless/session_test.rb
@@ -86,6 +86,14 @@ module Passwordless
       assert_equal "hi", session.token
       assert_equal Passwordless.digest("hi"), session.token_digest
     end
+    
+    test("setting token manually when case insensitive") do
+      Passwordless.config.case_insensitive_tokens = true
+      session = Session.new(token: "hi")
+      assert_equal "hi".upcase, session.token
+      assert_equal Passwordless.digest("hi".upcase), session.token_digest
+      Passwordless.config.case_insensitive_tokens = false
+    end
 
     test("with a custom expire at function") do
       custom_expire_at = Time.parse("01-01-2100").utc


README.md
```diff
@@ -296,6 +296,15 @@ end
 
 Passwordless will keep generating tokens until it finds one that hasn't been used yet. So be sure to use some kind of method where matches are unlikely.
 
+If your app should not distinguish between lowercase and uppercase letters in tokens, `Passwordless.config.case_insensitive_tokens` may be set to `true`.
+
+```ruby
+Passwordless.configure do |config|
+  config.token_generator = Passwordless::ShortTokenGenerator.new
```
bluefantail, line 303: Is this particular line relevant to case insensitivity example?

StevenClontz: Only because the default generator produces capitalized alphanumeric tokens. I'll make that clearer.

README.md
```diff
@@ -296,6 +296,15 @@ end
...
+  config.case_insensitive_tokens = true # allows `abc123` and `AbC123` to match `ABC123`
```
StevenClontz, line 304:
```suggestion
  config.token_generator = lambda do |_|
    [ *"A".."Z", *"0".."9" ].sample(8).join # produces eight-character alphanumeric tokens
  end
  config.case_insensitive_tokens = true # allows `abcd1234` and `AbCd1234` to match `ABCD1234`
```

README.md
StevenClontz: What do you think of the suggested change?

bluefantail: Oh I see — you're suggesting here as a hint for people using this that they may want to bump their token length to make up for the strength loss due to disabling case sensitive matching? I like this example showing those two parts together but wonder if that wants an even stronger callout tbh?

README.md
bluefantail: I know it's only two cherry picked examples lacking proper implementation context, but I had a look this afternoon for some examples — both Slack and Substack are using 6 digit codes.

Slack seem to be upcasing by default, Substack using only numbers. 6 Seems to be fairly normal-ish from a user experience POV? I couldn't speak to the rest of the variables in play with them though like expiry or token reuse etc.

On the flipside, Notion use much longer semi-memorable hyphen separated codes all in lower case like `som-fri-jaye-embol` just as a made up example.

I feel like this is a decision people need to make about security taking their specific context into account? It's fairly easy for a layperson like myself to just pick up this gem and do stuff with it which is what made me wonder if there should be a little bit of wording to that effect here  — or otherwise a confident recommendation? (which might be hard if there isn't a one size fits all approach)

mikker: I agree we should have a sensible default for most apps.

The threat here is someone brute forcing the password. Limiting factors
1. they need to know what email to target
2. we `BCrypt` on every guess, making every guess very slow

Just testing rn on my MacBook M1, just bcrypting one string takes around 250ms. Napkin math gives us about 2mil combinations `/[a-z0-9]/` (1,947,792).  Case-sensitive is about 62mil combinations (61,474,519). Significantly more – last factor is time:

3. How long before tokens time out

Default here is 10 mins ([#](https://github.com/mikker/passwordless/blob/ffa2d23e5ddbb87b4923d35fa68a2a443edc5469/lib/passwordless/config.rb#L38C1-L39C1)) so some more napkin math tells us a malicious actor can, without taking any other latency or restrictions into account, make 2400 guesses before the token times out.

My math might be wrong (?) but I think 6 is fine as a default if case-insensitive is already opt-in.

mikker: Perhaps we should add a "slowness" option like I think Devise has. Where you specify how many bcrypt runs a single try goes through. Every run halves the amount of tries one can do before timeout. 4 runs brings it down to 300 tries/10 mins and only adds 1sec of latency for a user, for example

bluefantail: That's a neat trick — would that be a case where each guess increases the runs by some amount IE ramping up the slowness after each failure? Or more of a hard parameter you can play with? I suppose it could take an optional function to let people do that sort of stuff themselves, though I guess that could also just be a bit of a footgun too.

All seems reasonable, unsure though what 'achievable guesses vs total combinations' ratio here is considered 'secure'?

Email I think is something reasonably attainable, whether it be a case of having been specifically targeted, or otherwise it could be more of an issue for really popular / large apps in terms of the sheer number of registered emails and resulting likelihood of finding one to work with, depending on your paranoia setting I guess :)

StevenClontz: If you're using the default `token_generator` https://github.com/mikker/passwordless/blob/ffa2d23e5ddbb87b4923d35fa68a2a443edc5469/lib/passwordless/short_token_generator.rb then there's no mixed case, and the case-insensitive feature doesn't make things any more/less secure (assuming the attacker knows tokens do not have a mix of upper/lower case).

I think this is a distraction, so I'll remove it.

StevenClontz, line 304:
```suggestion
```

StevenClontz:
```suggestion
  config.case_insensitive_tokens = true # allows `abc123` and `AbC123` to match `ABC123`
```

app/models/passwordless/session.rb
```diff
@@ -32,13 +32,15 @@ class Session < ApplicationRecord
     # hashed version in the database
     attr_reader :token
 
-    def token=(plaintext)
-      self.token_digest = Passwordless.digest(plaintext)
-      @token = (plaintext)
+    def token=(token)
+      Passwordless.config.case_insensitive_tokens ? modified_token = token.upcase : modified_token = token
+      self.token_digest = Passwordless.digest(modified_token)
+      @token = (modified_token)
```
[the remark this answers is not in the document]
StevenClontz:
```suggestion
    def token=(plaintext)
      self.token_digest = Passwordless.digest(plaintext)
      @token = (plaintext)
```

app/models/passwordless/session.rb
```diff
@@ -38,7 +38,8 @@ def token=(plaintext)
     end
 
     def authenticate(token)
-      token_digest == Passwordless.digest(token)
+      Passwordless.config.case_insensitive_tokens ? modified_token = token.upcase : modified_token = token
+      token_digest == Passwordless.digest(modified_token)
```
mikker, line 42:
```suggestion
      token = token.upcase if Passwordless.config.case_insensitive_tokens
      token_digest == Passwordless.digest(token)
```

mikker: personal preference

app/models/passwordless/session.rb
```diff
@@ -81,8 +82,9 @@ def set_defaults
 
       self.token, self.token_digest = loop {
         token = Passwordless.config.token_generator.call(self)
-        digest = Passwordless.digest(token)
-        break [token, digest] if token_digest_available?(digest)
+        Passwordless.config.case_insensitive_tokens ? modified_token = token.upcase : modified_token = token
+        digest = Passwordless.digest(modified_token)
+        break [modified_token, digest] if token_digest_available?(digest)
```
mikker, line 87:
```suggestion
        token = token.upcase if Passwordless.config.case_insensitive_tokens
        digest = Passwordless.digest(token)
        break [token, digest] if token_digest_available?(digest)
```

app/models/passwordless/session.rb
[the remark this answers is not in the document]
StevenClontz: I think we have to do this at write time - we're persisting the token digest, not the token, so I believe we cannot upcase the token digest when searching for tokens.

mikker (changes requested):

app/models/passwordless/session.rb
```diff
@@ -32,12 +32,14 @@ class Session < ApplicationRecord
...
-      self.token_digest = Passwordless.digest(plaintext)
-      @token = (plaintext)
+    def token=(token)
+      token = token.upcase if Passwordless.config.case_insensitive_tokens
+      self.token_digest = Passwordless.digest(token)
+      @token = token
     end
 
     def authenticate(token)
+      token = token.upcase if Passwordless.config.case_insensitive_tokens
```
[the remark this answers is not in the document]
StevenClontz: I'm not sure I follow. I believe this should work for a key generator that uses lower case keys (or mixed case, though that'd be silly). I'll add tests to confirm this when I have a chance.

Note that since the generator output is hashed before it is persisted to the database, we cannot store a mixed case token and hope to look it up with a case insensitive token.

Issue #241, opened 2025-01-07
Random ActiveRecord::RecordNotFound errors


noctivityinc:
The gem is working great most of the time.  However, I am catching a bunch of random errors like this:

```
ERROR TYPE	ActiveRecord::RecordNotFound
ERROR MESSAGE	Couldn't find Passwordless::Session with [WHERE "passwordless_sessions"."identifier" = $1 AND "passwordless_sessions"."authenticatable_type" = $2]
```

Any idea what would cause this and how I should handle them?   Here is the backtrace:

```
/gems/activerecord-7.2.2/lib/active_record/relation/finder_methods.rb:425 in raise_record_not_found_exception!
/gems/activerecord-7.2.2/lib/active_record/core.rb:319 in find_by!
/gems/passwordless-1.8.0/app/controllers/passwordless/sessions_controller.rb:252 in passwordless_session
/gems/passwordless-1.8.0/app/controllers/passwordless/sessions_controller.rb:54 in show
/gems/actionpack-7.2.2/lib/action_controller/metal/basic_implicit_render.rb:8 in send_action
/gems/actionpack-7.2.2/lib/abstract_controller/base.rb:226 in process_action
/gems/actionpack-7.2.2/lib/action_controller/metal/rendering.rb:193 in process_action
/gems/actionpack-7.2.2/lib/abstract_controller/callbacks.rb:261 in block in process_action
/gems/activesupport-7.2.2/lib/active_support/callbacks.rb:121 in block in run_callbacks
/gems/turbo-rails-2.0.11/lib/turbo-rails.rb:24 in with_request_id
/gems/turbo-rails-2.0.11/app/controllers/concerns/turbo/request_id_tracking.rb:10 in turbo_tracking_request_id
/gems/activesupport-7.2.2/lib/active_support/callbacks.rb:130 in block in run_callbacks
/gems/actiontext-7.2.2/lib/action_text/rendering.rb:25 in with_renderer
/gems/actiontext-7.2.2/lib/action_text/engine.rb:71 in block (4 levels) in <class:Engine>
/gems/activesupport-7.2.2/lib/active_support/callbacks.rb:130 in instance_exec
/gems/activesupport-7.2.2/lib/active_support/callbacks.rb:130 in block in run_callbacks
/gems/activesupport-7.2.2/lib/active_support/callbacks.rb:141 in run_callbacks
/gems/actionpack-7.2.2/lib/abstract_controller/callbacks.rb:260 in process_action
[...]

PR #262, opened 2025-10-15
Ignore "session not found" if paranoid is switched on

jamesjefferies:
If a bot or script is hitting the user sign in, then the `id` param can be something unsuitable, i.e. not an id in the database. The app then throws an `ActiveRecord::NotFound` exception.

This change alters the behaviour so that if paranoid is switched on, then the app ignores the not found and redirects back to the signin page.

However, I would understand if 

a) This should be the default behaviour and ignore the paranoid config
b) Current behaviour is desired!

This is related to issue #241

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -85,9 +85,13 @@ module Passwordless
 
       @session = passwordless_session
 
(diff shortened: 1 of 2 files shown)


mikker:
Thank you for this. Have you investigated what other auth solutions do around this?

jamesjefferies:
@mikker I can't say I have I'm afraid

PR #263, opened 2025-10-27
Expire session on sign out

mish-tan:
Had someone do a security check on our code base and they flagged this minor issue. 

Issue: sign_out does not invalidate the session in the db. If a bad actor gained access to the session cookie they would be able to log in as the user. 

To reproduce: 
1. Sign in, copy the cookie value for _session_id
2. Sign out 
3. Paste the cookie value back into _ session_id
4. Navigate to a page that only logged in users should see

This is because in our code base, we use `authenticate_by_session(authenticatable_class)` to verify the user is logged in, which checks if the session exists and it is available (not expired). If someone can get the old session_id, they can use it to log in because the old session is not expired. 

Fix: Expire the session when the user signs out, so it will not be available if hit again.

### DIFF
--- a/test/controllers/passwordless/sessions_controller_test.rb
+++ b/test/controllers/passwordless/sessions_controller_test.rb
@@ -273,7 +273,7 @@ module Passwordless
       assert_equal "/", path
       assert_match "Signed out successfully", flash[:notice]
       assert pwless_session(User).blank?
-      refute passwordless_session.reload.available?
+      assert_not passwordless_session.reload.available?
     end
 
     test("DELETE /:passwordless_for/sign_out :: When response options are configured ") do


mish-tan:
> Not immediately opposed to it. Have you looked if other libraries do this? Or the built in Rails auth generator?

(Apologies for the delayed response. Auth isn't an area I've had a lot of opportunity to touch before. I've had a look into some other libraries.) 

I think one of the things that makes the passwordless implementation a bit different is that the session is persisted in the DB, even after log out. I think a common thing to do is delete (and maybe reset) the session. 
> "Deleting the current_user_id key from the session is a typical way to log the user out" https://guides.rubyonrails.org/action_controller_overview.html#session

The built in Rails auth generator deletes the session on log out. 
https://github.com/rails/rails/issues/53129#issuecomment-2386511169

Devise appears to delete the session on logout 
https://github.com/heartcombo/devise/blob/c8a64b549c8b37e494eaca7be2def136a7e1b236/lib/devise/controllers/sign_in_out.rb#L71

"Clearance invalidates the users remember token and removes the remember token cookie from the store. "The invalidation of the remember token causes any other sessions that are signed in from other locations to also be invalidated on their next request."
https://github.com/thoughtbot/clearance/blob/14e9ba3d8f8bc865e389c743df4f9e8b2c7b3d9a/lib/clearance/session.rb#L70

mikker:
Thank you for the investigation. I think we should delete them but that's likely a breaking change. Feels like it at least. So let's do this and just expire them.

PR #264, opened 2025-11-07
[docs] Fix param order for after_session_confirm

timwis:
Hey @mikker hope you've been well. Just spotted that the example config options have the params reversed for `after_session_confirm`.


mikker:
Thanks!

PR #267, opened 2025-12-10
New.html.erb params.fetch always returns nil 

mish-tan:
Instead of using `params.fetch(email_field_name, nil)`, we should use `params.dig(:passwordless, email_field)`. 

If we use params.fetch(email_field_name), we try to look for something like this 
`params.fetch(:"passwordless[email]")`

But because rails converts params to a shape like this 
`passwordless: { email: "email@example.com" }`
`params.fetch(email_field_name, nil)` always returns nil.


mikker:
Thank you for this!

PR #269, opened 2026-03-06
Use flash.now instead of flash where we render 

mish-tan:
**Issue:** the alert from a failed login is appearing on other pages of our application (that can display flash alert/notice)

**Fix:** When we are calling render, we call `flash.now.alert =` instead of `flash.alert =`, and that will stop the error message from mistakenly appearing on other pages within the application. 

**Investigation Notes:**
According to here 
https://thoughtbot.com/blog/rails-flashes-guide#flash-vs-flashnow

(I also looked at here, but it's in Japanese. It has nicer screenshots tho. https://qiita.com/taraontara/items/2db82e6a0b528b06b949)

> When you set flash, the value is stored in session and persisted until the next request. This is useful for showing a message after a redirect.
> 
> When you set flash.now, the value is only persisted for the duration of the current request. This is useful for showing a message before rendering.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,8 @@ jobs:
             ruby: "3.0"
           - gemfile: gemfiles/Gemfile-rails-main
             ruby: "3.1"
+          - gemfile: gemfiles/Gemfile-rails-main
+            ruby: "3.2"
           - gemfile: gemfiles/Gemfile-rails-6-1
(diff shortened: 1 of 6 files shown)


mish-tan:
Sorry @mikker, I'm pretty confident in the code change, but I'm not sure where to update to get the checks to pass. Thanks for your time.

mikker:
Great! Thank you.

PR #271, opened 2026-06-18
Use "url_from" to validate destination_path to avoid potential open redirect attack 

### LINKED ISSUE #270: Security Issue, Redirect attack vulnerability
# Issue 

https://github.com/mikker/passwordless/issues/140

The change made for this PR means users could be redirected to other hosts, which is what Rails 7 tried to prevent. An attacker could manipulate the destination_path parameter to hijack the post-login redirect and send users to a malicious site.

e.g. when it's: `destination_path=-staging.com` . That would redirect to `https://staging.yoursite.net-staging.com`, which is off-site and thus an open redirect vulnerability.

## Proposed Fix 

Proposed way to fix this - we use `url_from` (https://github.com/rails/rails/blob/7a5951bd2aa321d9cc7c69358a9007797c97ccf1/actionpack/lib/action_controller/metal/redirecting.rb#L255) instead of the current validation of destination_path 

(Rails 8.1 introduced `allowed_redirect_hosts` https://github.com/rails/rails/pull/45680 so passwordless users don't need to set `allow_other_host: true` anymore, they can specify which hosts you should be a able to redirect a user to.)

### Current passwordless code:
```
    def passwordless_query_redirect_path
      query_redirect_uri = URI(params[:destination_path])
      query_redirect_uri.to_s if query_redirect_uri.host.nil? || query_redirect_uri.host == URI(request.url).host
    rescue URI::InvalidURIError, ArgumentError
      nil
    end
```

### Suggested Fix 
```
    def passwordless_query_redirect_path
      url_from(params[:destination_path])
    end

```

## How other gems handle this atm 

Devise parses the url (to redirect to) and extracts only the relative path before triggering the redirect. If users want to use allow_other_host: true they have to override the redirect_to method 

(https://github.com/heartcombo/devise/issues/5537, https://github.com/heartcombo/devise/pull/5462)

mish-tan:
Fixes [#270](https://github.com/mikker/passwordless/issues/270)

I updated the read me and added tests to confirm new behaviour. 

Users who are currently relying on both destination_path and allow_other_host will need to update their code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,21 @@ end
 
 This can also be turned off with `Passwordless.config.redirect_back_after_sign_in = false`.
 
+### Redirecting to external hosts (Multi-tenant applications)
+
+By default, `passwordless` strictly protects against Open Redirect vulnerabilities. When a user logs in, the `destination_path` parameter is sanitized using Rails' native `url_from` helper. This guarantees that users can only be redirected to safe, relative paths within your application.
+
+If you are building a multi-tenant application and legitimately need to redirect users to external subdomains or different hosts after authentication, you have two options depending on your Rails version:
+
+**For Rails 8.1+**
+Rails natively supports external redirect allowlisting. You can specify your trusted domains in your environment configuration, and `passwordless` will automatically respect them:
+
+# config/environments/production.rb
+config.action_controller.allowed_redirect_hosts = ["staging.yoursite.com", "app.yoursite.com"]
+
+**For Rails 8.0 and older**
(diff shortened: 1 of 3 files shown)


mish-tan:
CI is failing but it looks unrelated to my change... 

```
/home/runner/work/passwordless/passwordless/vendor/bundle/ruby/3.1.0/gems/i18n-1.15.0/lib/i18n.rb:58:in `config': undefined method `[]' for Fiber:Class (NoMethodError)

      current = Fiber[:i18n_config] || self.config = I18n::Config.new
```

